PR #6, opened 2024-03-18
Created empty project with Android Studio

octogradiste:


octogradiste:
Note that the Sonar analysis fails because we have 0% code coverage. We will fix this (i.e. add tests) once we start coding new features...

octogradiste:
Note that this will fix #2

.github/workflows/ci.yml
```diff
@@ -0,0 +1,115 @@
...
+    types:
+      - opened
+      - synchronize
+      - reopened
+
+jobs:
+  ci:
+    name: CI
+    # Execute the CI on the course's runners
+    runs-on: ubuntu-latest
```
violoncelloCH, line 19: I wonder if this still applies here since this repo is not part of the courses Github Organization and there are no runners configured for this project.
I'm not familiar with GitHub Actions, only with GitLab CI/CD and corresponding Runners... It's weird to me that this is running even though there are no Runners configured

violoncelloCH (approved):
Thanks, this looks good to me otherwise. 
Only remark: this was quite huge to go through, so in my opinion it would have made sense to split it up in multiple PRs (basing the further ones on top of the first which creates the base structure).

.github/workflows/ci.yml
octogradiste: I just checked the [docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#choosing-github-hosted-runners) and `ubuntu-latest` is a GitHub-hosted runner. We should remove the comment, it's just confusing...

violoncelloCH: Ohh, thanks for the clarification!

PR #7, opened 2024-03-18
add a description of the basic entities of the database

### LINKED ISSUE #5: data structures

unglazedstamp:
Basic description of the database for the app.
Created a mermaid diagram and a high level description of the entities.
fixes #5

### DIFF
--- a/DB_Structure.md
+++ b/DB_Structure.md
@@ -3,17 +3,22 @@
 This is the structure of the database.
 This structure only contains the basic elements needed for the main functions of the app.
 It is expected to evolve in the development of the app.
-- USER: a user of the app
-- EVENT: an event created in the app, created by a user
+- USER: a user of the app, metadata reference the attributes defined by https://supabase.com/docs/guides/auth/auth-user-management
+- USER_PROFILE: information related to a user
+- EVENT: an event created in the app, created by a user; the user has the choice to create the event for an association (host relation) or for himself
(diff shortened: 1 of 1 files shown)


octogradiste (changes requested):
I love the diagram!

Two things to think about:
1. Note that supabase automatically creates a user table. It contains User UID, Display Name, etc. but has no location attribute. I think we should store the locations of our users in a separate table, which would contain the User UID and the location. Maybe we can also drop the location for now and add it later once we implement this feature...
2. How do we know if an event is organized by an association or not ? I could be that a user is a committee member of an association but want's to create an event which is not tied to this association. Or the user is a member of multiple associations, to which one the event is linked to ? I think we can simply add to the event a FK to the association and then ensure that the organizer is actually a committee member of this association.

violoncelloCH:
> Maybe we can also drop the location for now and add it later once we implement this feature...

I agree, let's not include hypothetical things for now, but rather extend later on once we really need a field for a given feature.

unglazedstamp:
Thanks for the review,
1.  The user is now defined by supabase doc and the related information to a user are in user_profile. The location is not necessary for now, it's removed.
2. The idea is that the user has the choice when he creates an event to do it for an association or for himself. I added a host relation and a remark

octogradiste (approved):
Looks good to me!

PR #11, opened 2024-03-26
Feature/home-screen-composables

Greenade:
Changes to theme and color:
- adapted colors.kt with new colors for dark theme
- adapted theme.kt to always use dark theme (based on our app)
- adapted theme.kt to use new colors

Changes in main_activity:
- created the scaffold for the base of the screen, it contains the topbar only for now. but will hold floating buttons as well (to create events)
- implemented the top bar (the buttons are not interactable yet)
- implemented the content composable that will hold the main screen
- implemented the bottom sheet in the content composable, buttons aren't interactable yet
:warning:  need to be changed when implementing events on the map (open the bottom sheet only when event clicked), getting event info from database (to display in the bottom sheet) etc. :warning:

### DIFF
--- a/app/src/androidTest/java/com/github/swent/echo/compose/components/EventInfoSheetTest.kt
+++ b/app/src/androidTest/java/com/github/swent/echo/compose/components/EventInfoSheetTest.kt
@@ -1,0 +1,117 @@
+package com.github.swent.echo.compose.components
+
+import androidx.compose.ui.test.junit4.createComposeRule
+import androidx.compose.ui.test.onNodeWithTag
+import androidx.compose.ui.test.onNodeWithText
+import androidx.compose.ui.test.performClick
+import com.github.swent.echo.data.model.Event
+import com.github.swent.echo.data.model.Location
+import java.util.Date
+import org.hamcrest.CoreMatchers.equalTo
+import org.hamcrest.MatcherAssert.assertThat
+import org.junit.Rule
+import org.junit.Test
+
+class EventInfoSheetTest {
+
+    @get:Rule val composeTestRule = createComposeRule()
+
+    private var joinClicked = 0
+    private var peopleClicked = 0
+    private var dismissed = 0
+    private var extended = 0
+
+    private fun setUp(peopleMax: Int = 0) {
+        composeTestRule.setContent {
+            joinClicked = 0
+            peopleClicked = 0
+            dismissed = 0
+
+            val event =
+                Event(
+                    eventId = "1",
+                    organizerId = "1",
+                    title = "Event Title",
+                    description = "Event Description",
+                    location = Location("", 0.0, 0.0),
+                    startDate = Date(2024, 5, 18, 18, 15),
+                    endDate = Date(2024, 1, 1, 22, 0),
+                    tags = setOf()
+                )
+
+            EventInfoSheet(
+                event = event,
+                hostName = "Event Organization",
+                eventImage = android.R.drawable.ic_menu_help,
+                eventPeople = 0,
+                eventPeopleMax = peopleMax,
+                onJoinButtonPressed = { joinClicked++ },
+                onShowPeopleButtonPressed = { peopleClicked++ },
+                onDismiss = { dismissed++ },
+                onFullyExtended = { extended++ }
+            )
+        }
+    }
+
+    @Test
+    fun shouldShowEventTitle() {
+        setUp()
+        composeTestRule.onNodeWithText("Event Title").assertExists()
+    }
+
+    @Test
+    fun shouldShowEventDescription() {
+        setUp()
+        composeTestRule.onNodeWithText("Event Description").assertExists()
+    }
+
+    @Test
+    fun shouldShowEventDateTime() {
+        setUp()
+        composeTestRule.onNodeWithText("18/06\n" + "18:15").assertExists()
+    }
+
+    @Test
+    fun shouldShowEventOrganization() {
+        setUp()
+        composeTestRule.onNodeWithText("Event Organization").assertExists()
+    }
+
+    @Test
+    fun shouldShowJoinButton() {
+        setUp()
+        composeTestRule.onNodeWithText("Join Event").assertExists()
+    }
+
+    @Test
+    fun shouldShowPeopleButton() {
+        setUp()
+        composeTestRule.onNodeWithText("0").assertExists()
+    }
+
+    @Test
+    fun shouldCallJoinButtonPressedWhenJoinButtonClicked() {
+        setUp()
+        composeTestRule.onNodeWithTag("join_button").performClick()
+        assertThat(joinClicked, equalTo(1))
+    }
+
+    @Test
+    fun shouldCallShowPeopleButtonPressedWhenPeopleButtonClicked() {
+        setUp()
+        composeTestRule.onNodeWithTag("people_button").performClick()
+        assertThat(peopleClicked, equalTo(1))
+    }
+
+    @Test
+    fun shouldShowEventImage() {
+        setUp()
+        composeTestRule.onNodeWithTag("event_image").assertExists()
+    }
+
+    @Test
+    fun shouldShowPeopleCorrectlyWhenMaxPeopleIsNotZero() {
+        setUp(5)
+        composeTestRule.onNodeWithText("0/5").assertExists()
+    }
+}
--- a/app/src/androidTest/java/com/github/swent/echo/compose/components/TopBarTest.kt
+++ b/app/src/androidTest/java/com/github/swent/echo/compose/components/TopBarTest.kt
@@ -1,0 +1,43 @@
+package com.github.swent.echo.compose.components
+
+import androidx.compose.ui.test.junit4.createComposeRule
+import androidx.compose.ui.test.onNodeWithTag
+import androidx.compose.ui.test.onNodeWithText
+import androidx.compose.ui.test.performClick
+import org.hamcrest.CoreMatchers.equalTo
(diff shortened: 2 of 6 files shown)


octogradiste:
Files in the .idea/ folder are for configuring Android Studio. When I created the project, this folder was created automatically with all the necessary defaults. I pushed this folder, so that everybody has the default configuration. But I think we shouldn't push any new changes to this folder, otherwise it will mess up everybody's Android Studio configuration and create unnecessary merge conflicts. So, it would be cool if you could remove them from your PR.

octogradiste:
I like how the top bar and bottom sheet is looking! But do we really want to drop the light theme ? This is maybe something we should discuss on the next meeting... In general, I think defining the color scheme is something we should do later once our Figma is ready. Also we can then use this [tool](https://m3.material.io/theme-builder) to automatically generate a color theme.

octogradiste (changes requested):
You've put all your code in MainActivity. It would be cool if you could separate your composables into their own files. I suggest creating a package "compose" where you can put your composables. I would create a composable for the top bar and the bottom sheet. Then the person who develops the map screen will simply import them and add them to their scaffold. 

Also, if you separate your code in smaller composables it will be easier to test (we need 80% coverage).

app/src/main/java/com/github/swent/echo/compose/components/EventInfoSheet.kt
```diff
@@ -0,0 +1,195 @@
...
+    eventOrganization: String,
+    eventDescription: String,
+    eventDateTime: String,
+    eventImage: Int, // not sure how to handle images in the data system yet
+    eventPeople: Int,
+    eventPeopleMax: Int,
+    onJoinButtonPressed: () -> Unit,
+    onShowPeopleButtonPressed: () -> Unit
+) {
+    val sheetState = rememberModalBottomSheetState()
```
octogradiste, line 59: You can pass a `confirmValueChange` parameter to `rememberModalBottomSheetState()`, which is a callback you can use to determine if the bottom sheet is in the expanded state.

```kotlin
val sheetState = rememberModalBottomSheetState(skipPartiallyExpanded = false) {value ->
    if (value == SheetValue.Expanded) {
        // Call the onFullyExpanded callback
    }
    true
}
```

app/src/main/java/com/github/swent/echo/compose/components/EventInfoSheet.kt
```diff
@@ -0,0 +1,195 @@
...
+    // set to false initially, will need to change it to true to show the bottom sheet (by
+    // clicking events)
+    if (showBottomSheet.value) {
+
+        ModalBottomSheet(
```
octogradiste, line 64: Add a `Modifer.fillMaxSize()` so that it takes the full screen when dragged up.

app/src/main/java/com/github/swent/echo/compose/components/EventInfoSheet.kt
```diff
@@ -0,0 +1,195 @@
...
+            onDismissRequest = { showBottomSheet.value = false },
+            sheetState = sheetState,
+            containerColor = MaterialTheme.colorScheme.primaryContainer
+        ) {
+            // Sheet content
+            Box(
+                modifier =
+                    Modifier.padding(start = 20.dp, end = 20.dp, top = 0.dp, bottom = 32.dp)
+                        .width(360.dp)
+                        .height(375.dp)
```
octogradiste, line 74: Use `fillMaxWidth()` and `fillMaxHeight(0.5f)` to fill the bottom half of the screen. This is the `PartiallyExpanded` state of the `ModalBottomSheet`. Apply those before the padding.

octogradiste (changes requested):
Great work! The `TopBar` is really clean. For the `EventInfoSheet` I added some ideas on how to improve it. Let me know if you think they are not necessary or if there are some things which are not clear.

app/src/main/java/com/github/swent/echo/compose/components/EventInfoSheet.kt
```diff
@@ -0,0 +1,195 @@
...
+    eventTitle: String,
+    eventOrganization: String,
+    eventDescription: String,
+    eventDateTime: String,
+    eventImage: Int, // not sure how to handle images in the data system yet
+    eventPeople: Int,
+    eventPeopleMax: Int,
+    onJoinButtonPressed: () -> Unit,
+    onShowPeopleButtonPressed: () -> Unit
+) {
```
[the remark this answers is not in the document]
Greenade: yeah, onDismissed makes sense if we don't handle the visibility of the sheet in itself, for onFullyExpanded, I don't know if we can do that... I will look into it

app/src/main/java/com/github/swent/echo/compose/components/EventInfoSheet.kt
Greenade: I'll try that !

app/src/main/java/com/github/swent/echo/compose/components/EventInfoSheet.kt
Greenade: that makes sense, I'll change that

app/src/main/java/com/github/swent/echo/compose/components/EventInfoSheet.kt
```diff
@@ -0,0 +1,195 @@
...
+@Composable
+fun EventInfoSheet(
+    showBottomSheet: MutableState<Boolean>,
+    eventTitle: String,
+    eventOrganization: String,
+    eventDescription: String,
+    eventDateTime: String,
+    eventImage: Int, // not sure how to handle images in the data system yet
+    eventPeople: Int,
+    eventPeopleMax: Int,
```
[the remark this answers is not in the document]
Greenade: the people should not be an attribute I think, but a value we get by counting the number of entries in the relation between Event and People (on the selected event) but for the rest yes they should be added I think, I don't know how exactly we should handle images in the database though, because we can't just store every image locally and refer to them, that doesn't make any sense...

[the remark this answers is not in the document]
violoncelloCH: yes, so far I don't touch the attributes of the individual objects, so all good for modifying the Event class...
however I'm not sure yet, for interacting with the database I might need to do it...but that's an other discussion

for the images I don't know yet - we'll have to see how this has to be done properly...
the images will be stored in the supabase storage (not the database) and cached locally on the device... the database only holds a reference to the image

octogradiste (dismissed):
Looks excellent, I think we should merge this into the main branch so we can start using it! We will still be able to change the function signature of the `EventInfoSheet` later on...

app/src/androidTest/java/com/github/swent/echo/compose/components/TopBarTest.kt
```diff
@@ -0,0 +1,41 @@
...
+
+    @Test
+    fun shouldShowEcho() {
+        composeTestRule.onNodeWithText("Echo").assertExists()
+    }
+
+    @Test
+    fun shouldCallOpenMenuCallbackWhenMenuButtonClicked() {
+        composeTestRule.onNodeWithTag("menu_button").performClick()
+        assert(menuClicked == 1)
```
david-schroeter, line 33: The  `JVM assert()` method isn't really meant to be used in test. It should be used directly in the code when some property needs to be asserted.

In tests a test framework should be preferred, that will provide more functionality, more advanced matchers and more explicit failure messages.

It will often better describe the test that you perform.

For example by using Hamcrest (which is present by default with junit in android) it could be written that way: `assertThat(menuClicked, equalTo(1))`

Same for all other `assert()` in other tests.

app/src/androidTest/java/com/github/swent/echo/compose/components/EventInfoSheetTest.kt
```diff
@@ -0,0 +1,137 @@
...
+
+    @Test
+    fun shouldShowEventImage() {
+        setUp()
+        composeTestRule.onNodeWithTag("event_image").assertExists()
+    }
+
+    @Test
+    fun shouldShowPeopleCorrectlyWhenMaxPeopleIsNotZero() {
+        // not using setUp() because we need to set eventPeopleMax to 5
```
david-schroeter, line 111: Wouldn't it be possible to add a parameter to setup ?

`private fun setUp(eventPeopleMax: Int = 0) {  ... }` and then call `setUp(5)`

david-schroeter (commented):
Great work, the structure of your code is very good, and the code is clean!

I just added a couple of comments that may help you in the future.

But overall very good PR!

app/src/main/java/com/github/swent/echo/compose/components/TopBar.kt
```diff
@@ -0,0 +1,71 @@
...
+    // Top app bar with title, navigation icon, and actions
+    CenterAlignedTopAppBar(
+        colors =
+            TopAppBarDefaults.topAppBarColors(
+                containerColor = MaterialTheme.colorScheme.primaryContainer,
+                titleContentColor = MaterialTheme.colorScheme.onPrimary,
+            ),
+        title = {
+            Text(
+                "Echo",
```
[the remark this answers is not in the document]
Greenade: I will look into it 👍

Greenade:
The number of commits here is not normal and not intentional :( When I rebase I just need to add everything back it's weird, after the last review I only changed the tests

violoncelloCH (approved):
thanks for addressing the points; looks all good to me now :)

PR #15, opened 2024-03-27
Feature/navigation

unglazedstamp:
Navigation routes, actions and main composable. Should be ready to handle navigation actions between screens.


alejandrocalles (approved):
Good job! Everything looks clean and easy to extend with more destinations.

app/build.gradle.kts
```diff
@@ -53,6 +53,15 @@ android {
...
             excludes += "/META-INF/{AL2.0,LGPL2.1}"
+            excludes += "META-INF/LICENSE.md"
+            excludes += "META-INF/LICENSE-notice.md"
+        }
+    }
+    testOptions {
+        packagingOptions {
+            jniLibs {
+                useLegacyPackaging = true
+            }
```
octogradiste, line 64: This is already in the main branch, maybe you can roll back those changes and then rebase main.

app/build.gradle.kts
```diff
@@ -86,6 +99,8 @@ dependencies {
     androidTestImplementation(libs.androidx.espresso.core)
     androidTestImplementation(platform(libs.androidx.compose.bom))
     androidTestImplementation(libs.androidx.ui.test.junit4)
+    androidTestImplementation(libs.androidx.navigation.testing)
+    androidTestImplementation(libs.mockk.android)
```
octogradiste, line 103: Note that mockk is already an android test dependency in main.

app/src/main/java/com/github/swent/echo/ui/navigation/Navigation.kt
```diff
@@ -0,0 +1,60 @@
...
+    SIGN_IN,
+    MAP,
+}
+
+/**
+ * This class reprensent the different navigation actions
+ *
+ * @param navController the navigation controller
+ */
+class navigationActions(val navController: NavHostController) {
```
octogradiste, line 23: Nitpicking: The convention is to start class names with capital letters: `NavigationActions`. :)

app/src/main/java/com/github/swent/echo/ui/navigation/Navigation.kt
```diff
@@ -0,0 +1,60 @@
...
+}
+
+/**
+ * Navigation composable, it display the relevant screen based on the route and pass an instance of
+ * navigationActions to the screens
+ *
+ * @param navController the navigation controller, created if none is given
+ */
+@Composable
+fun appNavigationHost(
```
octogradiste, line 43: Nitpicking: The convention is to start composables with a capital letter: `AppNavigationHost`.
Nitpicking: I think it could make sense to extract this into its own file.

octogradiste (changes requested):
Looks really clean, I like it a lot! 
Just make sure that rebasing doesn't mess up the dependencies, as some of them are now already in the main branch...
And sorry, I love nitpicking ;)

app/build.gradle.kts
violoncelloCH: no need to roll back, just doing the rebase should do the work :) either the changes are exactly are exactly identical and git will understand magically that it's already in main; otherwise it will let you resolve the conflict...

violoncelloCH:
What happened here, why are there suddenly 89 commits on this branch?
Ohh, looks like a merge instead of a rebase, right? This really messes up the PR since there are a lot of unrelated changes shown here now (more than 1000 lines, a lot of them not related to what you implemented if y recall correctly)... so even for reviewing it makes it really hard. Could you do a proper rebase, please? If you want some assistance I'm happy to hop through the rebase together with you :)

violoncelloCH (changes requested):
putting this in as a blocker, see comment above ^

app/build.gradle.kts
unglazedstamp: Solved with the rebase

violoncelloCH (approved):

app/build.gradle.kts
unglazedstamp: Solved with the rebase

app/src/main/java/com/github/swent/echo/ui/navigation/Navigation.kt
unglazedstamp: done

app/src/main/java/com/github/swent/echo/ui/navigation/Navigation.kt
unglazedstamp: done

unglazedstamp:
Sorry for the inconvenience, I used some git features of Android Studio without checking the result. It should be a proper rebase now. Thanks for the quick review and proposition of help.

violoncelloCH (approved):
thanks for doing a clean rebase and for implementing the changes suggested by @octogradiste ! looks all good now!

Issue #9, opened 2024-03-22
Home screen map


alejandrocalles:
Create a Composable with the map, wtih scroll and zoom functionality.
Add a way to create and drop pins on the map.
Add tests.

PR #17, opened 2024-03-28
Feature/map composable

alejandrocalles:
Addresses issue #9. Add a map composable using the `osmdroid` library. Should be able to load a map correctly. The map tiles should be cached so that the map is still displayed (within reasonable limits) when the app is offline. The `MapView` should be able to display custom markers.


violoncelloCH (approved):
LGTM, tested and works!
thanks for the work...
one suggestion: I just stumbled over https://docs.maptiler.com/maplibre-gl-native-android/ which seems to be the more modern way to embed osm into an android app. It uses vector tiles and the style is also more modern than the default osm tiles. What do you think, could we look into this? Should I add it to the backlog instead?

alejandrocalles:
Thanks! I'll look into MapLibre. It might be worth the hassle as it looks much better.

violoncelloCH:
Cool! I'll create an issue to track this :)

PR #26, opened 2024-03-28
Hilt integration

### LINKED ISSUE #25: integrate hilt

octogradiste:
This will close #25.


alejandrocalles (changes requested):
Really great job! Both the code and the directory structure are clean and seem well thought of. The only small detail I would say needs to be changed that the supabase public key should be treated as a resource rather than a hardcoded value. The main reason is that it becomes easier to change the key afterwards (maybe even use a separate key for a testing environment if we ever deploy our app).
One way to do this would be to write the public key to a file in `app/src/main/res/raw/`. The `raw` directory doesn't seem to exist for the moment but you can always create it. You can then load the public key from said file. From what I've read online, the `androidTest` code should also be able to find the public key at the same place.

octogradiste:
Thanks very much for your review! When I hard-coded those strings, I actually thought of putting them into the string resources. I quickly gave up as I thought I need the context of the `Activity`, which I don't have access to in a `SingletonComponent`. But your review motivated me to look a bit deeper.  I found out that the `SingletonComponent` is actually bound to the lifecycle of the `Application` which also has a context! Using this context, I was able to access the resources :)
By the way, I placed the strings into the `values/strings.xml` instead of a file in the `raw` folder, as I feel like this makes more sense...

alejandrocalles (approved):
I'm glad you managed to solve it! Apologies for sending you in a deep dive. In any case, the code for `SupabaseClientModule` looks really solid now. Thanks a lot!

PR #30, opened 2024-04-02
repository pattern interfaces

### LINKED ISSUE #12: implement repository pattern
This [blog](https://medium.com/@dugguRK/the-real-repository-pattern-in-android-refers-to-an-architectural-approach-that-abstracts-the-fd4f4cc573c3#:~:text=The%20%E2%80%9CReal%E2%80%9D%20Repository%20Pattern%20in,architecture%2C%20testability%2C%20and%20maintainability) post explains the repository pattern. The goal is to simply define a repository and datasource interface which can be injected into the viewmodels. The implementations of those interfaces will be done later on.

violoncelloCH:
resolves #12
those are only the interfaces related to the repository pattern
the implementation will follow in a later PR but depends on the actual data sources for which the implementation will take longer

Note that LocalDataStorage and RemoteDataStorage currently contain no actual methods. This is intentional since they will need to be adapted to the data sources and the synchronisation mechanism.

@echo-swent/group09 If you already know that you'll need a specific setter or getter for your view / UI code, please comment your exact need or add them here. The idea is that we can get that merged as soon as possible.

### DIFF
--- a/app/src/main/java/com/github/swent/echo/data/repository/Repository.kt
+++ b/app/src/main/java/com/github/swent/echo/data/repository/Repository.kt
@@ -1,0 +1,35 @@
+package com.github.swent.echo.data.repository
+
+import com.github.swent.echo.data.model.Association
+import com.github.swent.echo.data.model.Event
+import com.github.swent.echo.data.model.Tag
+import com.github.swent.echo.data.model.User
+import com.github.swent.echo.data.model.UserProfile
+
+interface Repository {
+    suspend fun getAssociation(associationId: String): Association
+
+    suspend fun setAssociation(association: Association)
+
+    suspend fun getAllAssociations(): List<Association>
+
+    suspend fun getEvent(eventId: String): Event
+
+    suspend fun setEvent(event: Event)
+
+    suspend fun getAllEvents(): List<Event>
+
+    suspend fun getTag(tagId: String): Tag
+
+    suspend fun setTag(tag: Tag)
+
+    suspend fun getAllTags(): List<Tag>
+
+    suspend fun getUser(userId: String): User
+
+    suspend fun setUser(user: User)
+
+    suspend fun getUserProfile(userId: String): UserProfile
+
+    suspend fun setUserProfile(userProfile: UserProfile)
+}
--- a/app/src/main/java/com/github/swent/echo/data/repository/datasources/LocalDataSource.kt
+++ b/app/src/main/java/com/github/swent/echo/data/repository/datasources/LocalDataSource.kt
@@ -1,0 +1,5 @@
+package com.github.swent.echo.data.repository.datasources
+
+interface LocalDataSource {
+    fun getData(): String
+}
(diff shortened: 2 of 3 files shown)


Greenade (commented):
Good interface, I think we still need some getters for relations, meaning for example registrationEvent(event: Event): Int / registrationEvent(event: Event): List<UserProfile> to see who is registrered in a certain event (or simply how many people). It would also be useful to be able to get a list of tags concerning a certain event, or the list of tags that a User is interested in. Also to be able to get the host of an Event if there is any would be needed (get the Association). What I mentioned here is non exhaustive but you get the idea.
Other than that the images of the Events should also be a separate getter in my opinion [getImage(event: Event): Image] Since we want to cache them and not have them all in the local files at all times. (not sure about images again, but i still mention it just in case)

octogradiste (changes requested):
Thank you for working on the Repository!!!

I think the `getType()`'s should take an `typeId` as argument, otherwise, which `Type` are you returning? The use case of those getters would be to get an updated version of this `typeId`. Example: on the map we show all events, once the user clicks on an event we can call the `getEvent(eventId)` to get the newest version.

Note that the interfaces for the local and remote data source are not important yet, as they are only going to be used by you (or the one implementing the `Repository`). For the view models, we only need the `Repository` interface.

Also, for now, the `Repository` could just expose one function for getting all events and then the view model handles the filtering. So, you don't have to expose a function for every possible filtering option. Maybe, in the future when we have millions of events we could add functions which would run SQL queries for better performance :) 

One more thing, I would mark all the functions (or at least the getters) as `suspend`, as they will run asynchronous code (query local and remote database).

violoncelloCH:
Hi, thanks the two of you for your reviews!

> I think the getType()'s should take an typeId as argument, otherwise, which Type are you returning? The use case of those getters would be to get an updated version of this typeId. Example: on the map we show all events, once the user clicks on an event we can call the getEvent(eventId) to get the newest version.

you're right, I completely missed that -> I've updated the PR accordingly

> Also, for now, the Repository could just expose one function for getting all events and then the view model handles the filtering. So, you don't have to expose a function for every possible filtering option. Maybe, in the future when we have millions of events we could add functions which would run SQL queries for better performance :)

I guess that makes sense! for now let's focus on the essential stuff and implement sophisticated getters later when we know exactly what we need in the ViewModel... I'll only have the basic setters here - we can extend that later...

> One more thing, I would mark all the functions (or at least the getters) as suspend, as they will run asynchronous code (query local and remote database).

thanks! added that as well :)

@Greenade : in accordance with what @octogradiste proposed, I think we should go with just the bare minimum here and extend later when we see where it makes sense to optimize for performance.
The host of an event can be fetched using the organizerId in th Event. The signups however are not even reflected yet in the database scheme - I guess that needs an update. Possibly another table holding the signups, relating user to event, would be needed for this

Greenade:
Can we not implement the things I talked about with the help of the relation tables ? I mean a SQL query seems doable (count * from participant where eventid = eventid) something like that ?

violoncelloCH:
@Greenade yes we will implement them, but in a second round. For the start we need a minimal base here. Implementing the whole repository pattern and access of the data sources is a huge job already as is. The idea is to get a minimal implementation implemented first; then we can easily add on top of that and implement more sophisticated getters and setters which do more complex queries in an iterative process.

octogradiste (changes requested):
Thanks for adding the changes!! I'd suggest renaming the interface to `Repository` as suggested in #33.

octogradiste (approved):
Looks good to me! Thanks so much for working on this!

PR #31, opened 2024-04-02
room DB for local caching

violoncelloCH:
depends on #32 

this is only a start for now
it will serve as a backend for the LocalDataSource of the Repository

room documentation: https://developer.android.com/training/data-storage/room
how to construct relations in room: https://developer.android.com/training/data-storage/room/relationships

### DIFF
--- a/app/src/main/java/com/github/swent/echo/data/room/RoomLocalDataSource.kt
+++ b/app/src/main/java/com/github/swent/echo/data/room/RoomLocalDataSource.kt
@@ -47,10 +47,10 @@ class RoomLocalDataSource @Inject constructor(db: AppDatabase) : LocalDataSource
     override suspend fun setEvent(event: Event) {
         val crossRefs = event.tags.map { EventTagCrossRef(event.eventId, it.tagId) }
 
-        eventDao.insert(EventRoom(event))
-        eventDao.insertEventTagCrossRegs(crossRefs)
         event.organizer?.let { associationDao.insert(AssociationRoom(it)) }
         tagDao.insertAll(event.tags.toTagRoomList())
+        eventDao.insertEventTagCrossRegs(crossRefs)
+        eventDao.insert(EventRoom(event))
     }
 
     override suspend fun getAllEvents(): List<Event> {
@@ -63,10 +63,10 @@ class RoomLocalDataSource @Inject constructor(db: AppDatabase) : LocalDataSource
         val crossRefs =
             events.flatMap { event -> event.tags.map { EventTagCrossRef(event.eventId, it.tagId) } }
 
-        eventDao.insertAll(events.map { EventRoom(it) })
-        eventDao.insertEventTagCrossRegs(crossRefs)
         associationDao.insertAll(associations.map { AssociationRoom(it) })
         tagDao.insertAll(tags.toTagRoomList())
+        eventDao.insertEventTagCrossRegs(crossRefs)
+        eventDao.insertAll(events.map { EventRoom(it) })
     }
 
     override suspend fun getTag(tagId: String): Tag? {
@@ -97,8 +97,8 @@ class RoomLocalDataSource @Inject constructor(db: AppDatabase) : LocalDataSource
         val tags = userProfile.tags.toTagRoomList()
         val crossRefs = tags.map { UserProfileTagCrossRef(userProfile.userId, it.tagId) }
 
-        userProfileDao.insert(UserProfileRoom(userProfile))
-        userProfileDao.insertUserProfileTagCrossRefs(crossRefs)
         tagDao.insertAll(tags)
+        userProfileDao.insertUserProfileTagCrossRefs(crossRefs)
+        userProfileDao.insert(UserProfileRoom(userProfile))
     }
 }
--- a/app/src/main/java/com/github/swent/echo/data/room/RoomLocalDataSource.kt
+++ b/app/src/main/java/com/github/swent/echo/data/room/RoomLocalDataSource.kt
@@ -98,7 +98,7 @@ class RoomLocalDataSource @Inject constructor(db: AppDatabase) : LocalDataSource
         val crossRefs = tags.map { UserProfileTagCrossRef(userProfile.userId, it.tagId) }
 
         userProfileDao.insert(UserProfileRoom(userProfile))
-        userProfileDao.insertUserProfileTagCrossRegs(crossRefs)
+        userProfileDao.insertUserProfileTagCrossRefs(crossRefs)
         tagDao.insertAll(tags)
     }
 }
--- a/app/src/main/java/com/github/swent/echo/data/room/RoomLocalDataSource.kt
+++ b/app/src/main/java/com/github/swent/echo/data/room/RoomLocalDataSource.kt
@@ -9,8 +9,11 @@ import com.github.swent.echo.data.room.entity.AssociationRoom
 import com.github.swent.echo.data.room.entity.EventRoom
 import com.github.swent.echo.data.room.entity.EventTagCrossRef
 import com.github.swent.echo.data.room.entity.TagRoom
+import com.github.swent.echo.data.room.entity.UserProfileAssociationSubscriptionCrossRef
+import com.github.swent.echo.data.room.entity.UserProfileCommitteeMemberCrossRef
 import com.github.swent.echo.data.room.entity.UserProfileRoom
 import com.github.swent.echo.data.room.entity.UserProfileTagCrossRef
+import com.github.swent.echo.data.room.entity.toAssociationRoomList
 import com.github.swent.echo.data.room.entity.toTagList
 import com.github.swent.echo.data.room.entity.toTagRoomList
 import javax.inject.Inject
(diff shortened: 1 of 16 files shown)


violoncelloCH:
@srsingh04 no, I don't think so... I'm not even sure if it makes sense to have that as an attribute on the data class, but whether a tag has been selected in the UI is state only concerning the UI. It doesn't make sense to store that in DB and especially not on Supabase since that would mean it would get synchronized between all users.

app/src/main/java/com/github/swent/echo/data/repository/datasources/Cache.kt
```diff
@@ -0,0 +1,21 @@
+package com.github.swent.echo.data.repository.datasources
+
+import android.app.Application
+import androidx.room.Room
+import com.github.swent.echo.data.room.AppDatabase
+import com.github.swent.echo.data.room.Converters
+
+class Cache(application: Application) : LocalDataSource {
```
violoncelloCH, line 8: I'd suggest to rename this to `RoomDataSource` and move it to the room/ dicrectory/package for it to be consistent :)

app/src/main/java/com/github/swent/echo/data/room/entity/EventRoom.kt
```diff
@@ -0,0 +1,79 @@
...
+import androidx.room.Junction
+import androidx.room.PrimaryKey
+import androidx.room.Relation
+import com.github.swent.echo.data.model.Event
+import java.time.ZonedDateTime
+
+@Entity
+data class EventRoom(
+    @PrimaryKey val eventId: String,
+    @Embedded val creator: EventCreatorRoom,
```
violoncelloCH, line 15: If I understand correctly how `@Embedded` works, this will add the fields of `EventCreatorRoom` to the Event table. I'm not sure if this is actually what we want. The event creator should be a subset of the attributes of `userProfile` and then it could be a DB relation between `UserProfile` and `Event`. Otherwise changes to the `UserProfile` won't be visible on the events.

Maybe [views](https://developer.android.com/training/data-storage/room/creating-views) could help us implement this properly? However they only allow to do "SELECT" queries. 
Also the downside of this approach would be that we'd either have a situation where `userProfile`s would be incomplete in the local database (i.e. only the attributes part of `CreatorProfile` have been loaded from the remote). Or we'd need to fetch the complete the compete `UserProfile`s as soon as we have an Event referencing it (not what we want either).

What do you think makes the most sense here? Would the "incomplete `UserProfile`s" approach make sense to you?

app/src/main/java/com/github/swent/echo/data/room/entity/UserProfileRoom.kt
```diff
@@ -0,0 +1,53 @@
...
+import com.github.swent.echo.data.model.UserProfile
+import com.github.swent.echo.data.model.toSectionEPFL
+import com.github.swent.echo.data.model.toSemesterEPFL
+
+@Entity
+data class UserProfileRoom(
+    @PrimaryKey val userId: String,
+    val name: String,
+    val semester: String?,
+    val section: String?,
```
violoncelloCH, line 18: I already started adding association committee memberships as well as association following as relations between `UserProfile` and `Association` over at #157. So this might be something you already want to take a look at depending on which PR gets merged first.

violoncelloCH (commented):
Thanks for working on the local DB implementation :) It is really well done in general! 
I have some specific questions and feedback though in the comments below.

app/src/main/java/com/github/swent/echo/data/room/entity/EventRoom.kt
```diff
@@ -0,0 +1,79 @@
...
+        EventCreatorRoom(event.creator),
+        event.organizer?.associationId,
+        event.title,
+        event.description,
+        LocationRoom(event.location),
+        event.startDate,
+        event.endDate,
+        event.participantCount,
+        event.maxParticipants,
+        null,
```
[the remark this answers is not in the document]
octogradiste: This is something we should discuss in the team, but as I understood, we want to drop the images as it doesn't add much and adds a lot of complexity. Also, we store an `imageId` (an `int`) inside the `Event`, if we want to use images, we probably should use a `String` to store the path of the image. That why I store a `imagePath` inside the `EventRoom`. But, as I said, this is something we should discuss with the team during the next meeting.

app/src/main/java/com/github/swent/echo/data/room/dao/UserProfileRoomDao.kt
```diff
@@ -0,0 +1,23 @@
...
+import com.github.swent.echo.data.room.entity.UserProfileTagCrossRef
+import com.github.swent.echo.data.room.entity.UserProfileWithTags
+
+@Dao
+interface UserProfileRoomDao {
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    suspend fun insert(userProfile: UserProfileRoom)
+
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    suspend fun insertUserProfileTagCrossRegs(userProfileTagCrossRef: List<UserProfileTagCrossRef>)
```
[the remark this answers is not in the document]
octogradiste: Good catch, I should correct this!

app/src/main/java/com/github/swent/echo/data/room/RoomLocalDataSource.kt
```diff
@@ -0,0 +1,104 @@
...
+        return eventDao.get(eventId)?.toEvent()
+    }
+
+    override suspend fun setEvent(event: Event) {
+        val crossRefs = event.tags.map { EventTagCrossRef(event.eventId, it.tagId) }
+
+        eventDao.insert(EventRoom(event))
+        eventDao.insertEventTagCrossRegs(crossRefs)
+        event.organizer?.let { associationDao.insert(AssociationRoom(it)) }
+        tagDao.insertAll(event.tags.toTagRoomList())
```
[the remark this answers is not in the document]
octogradiste: Good point, I'll change that just in case, but as all tests pass (also the one where I set an event with tags and then retrieve it back), I feel like this isn't really an issue when using room.

app/src/main/java/com/github/swent/echo/data/room/entity/EventRoom.kt
octogradiste: Yeah, you're right, it's not the best solution. My thinking was that the name of the user won't change often. Therefore, I thought it's acceptable that in some rare cases, it might be out of sync with the real user profile name. Also, we are using the local data source only for caching and not as a source of truth. We could add some mechanism in the UI (e.g. pull down to refresh) to update/sync/refresh the data...
Otherwise, we could just not offer the possibility to the user to change his name... this we might also discuss with the team.

app/src/main/java/com/github/swent/echo/data/room/entity/UserProfileRoom.kt
octogradiste: OK, thanks for the info, I'll take a look!
By the way, thank you so much for taking the time to review all this!!!

app/src/main/java/com/github/swent/echo/data/room/entity/EventRoom.kt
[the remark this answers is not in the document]
violoncelloCH: you're right that storing an `int` is not the right thing to do... but I think it makes sense to already prepare this DB column for later once we want to add that feature. So changing that to storing a filepath or some kind of resource identifier probably makes sense... I'll take a look how storing blobs on supabase works.

app/src/main/java/com/github/swent/echo/data/room/entity/EventRoom.kt
violoncelloCH: Agree that that is a low priority thing and as you say it's not the source of truth it's acceptable to be out of sync. Offering a pull-to-refresh to force-refresh is a good idea! We should look into this.
Treating the names (or similar attributes) as immutable however is generally a bad idea. Users might simply misspell it on profile creation and want to correct it or want to change names for a diverse range of reason (marriage, gender, identity, etc)...

octogradiste: Yes, you're right, we should let the user the freedom to change their name. But, as you said, this is not high priority, so we should leave this for the future and keep this solution for now!

app/src/main/java/com/github/swent/echo/data/room/entity/EventRoom.kt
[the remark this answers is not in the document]
octogradiste: I create an issue #165 in case we want to add images later on...

app/src/main/java/com/github/swent/echo/data/room/entity/EventRoom.kt
octogradiste: I opened an issue #166 for the pull-to-refresh support.

octogradiste (dismissed):
@violoncelloCH can't approve because he created the PR, so I have to do it for him ;)

octogradiste:
We might want to wait until #157 is merged. This way, I can adapt the local database to the new changes of the `Event` data class...

unglazedstamp (approved):
Approved by @violoncelloCH

PR #32, opened 2024-04-02
repository implementation

violoncelloCH:
~requires #30 to be merged first~ done

This implements a first version of the repository and a supabase connector.
In this first version, data is directly fetched from supabase. Caching will be implemented later, see #31 and #50.
There are also other changes we need to do to improve the datamodel and how to access data. For instance, getting an event is not yet working since we need to correctly fetch the associated tags: #68; or also: #67


unglazedstamp (approved):
The code looks very clean !
Thank you for implementing the repository !

PR #37, opened 2024-04-03
Improve data classes `Event` and `Location`.

### LINKED ISSUE #36: Use a GeoPoint for the location.

alejandrocalles:
Closes #36. Add utility function `toGeoPoint` to `Location` data class. Add secondary constructor to `Location` data class. Add an immutable `Set` of `Tag`s as a field of the `Event` data class.

### DIFF
--- a/app/src/main/java/com/github/swent/echo/data/model/Event.kt
+++ b/app/src/main/java/com/github/swent/echo/data/model/Event.kt
@@ -9,5 +9,6 @@ data class Event(
     val description: String,
     val location: Location,
     val startDate: Date,
-    val endDate: Date
+    val endDate: Date,
+    val tags: Set<Tag>
 )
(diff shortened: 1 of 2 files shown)


Greenade (commented):
do we need an immutable set of tags as a field of an event ? Seems to me it would make more sense to access that with an SQL request based on the relation between the Event class and the Tag class... It would make it possible to add tags and remove them as well.

violoncelloCH (approved):
nice addition! LGTM :)

violoncelloCH:
> do we need an immutable set of tags as a field of an event ? Seems to me it would make more sense to access that with an SQL request based on the relation between the Event class and the Tag class... It would make it possible to add tags and remove them as well.

I think that having that as a List directly available on the event makes sense here. Keep in mind that this is just the data class. It does not have anything to do with how things are stored in the database. Having the tags as a list here just means that the job of fetching the tags from the database will need to be done by the repository/data source and not the view model. Having immutability on the data class also does not imply that an event could not be modified later on. Modifications can be done by creating a copy of the object with your modifications and sending it back to the repository through the respective setter. The repository then updates the respective fields of the entities or the relations in the database(s)

PR #38, opened 2024-04-04
Redesign the way MapViews are created to add support for MapLibre.

### LINKED ISSUE #34: use Mapblibre SDK instead of osmdroid
While reviewing #17 I stumbled over the Maplibre SDK which seems to be the more modern way to embed osm into an android app. It uses vector tiles and the style is also more modern than the default osm tiles. 

https://docs.maptiler.com/maplibre-gl-native-android/

### LINKED ISSUE #35: Add support for MapLibre.
Add an interface for Compose to interact with MapViews.
Add a class for `osmdroid` and another for `MapLibre`.
Add the necessary tests.

alejandrocalles:
Will close #34. The objective is to abstract the way `MapView`s are provided in order to be able to choose freely between `osmdroid` and `MapLibre` view providers.
Ideally, support for markers will also be added for both `osmdroid` and `MapLibre`.

### DIFF
--- a/app/src/main/java/com/github/swent/echo/compose/map/Map.kt
+++ b/app/src/main/java/com/github/swent/echo/compose/map/Map.kt
@@ -1,28 +1,62 @@
 package com.github.swent.echo.compose.map
 
+import android.content.Context
 import android.view.View
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.MutableState
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.viewinterop.AndroidView
-import org.osmdroid.util.GeoPoint
-import org.osmdroid.views.MapView
-
-fun updateMapView(view: MapView, newCenter: GeoPoint) {
-    view.apply { controller.setCenter(newCenter) }
-}
+import com.github.swent.echo.data.model.Event
 
 @Composable
-fun <T : View> MapDrawer(
+fun <T : View> EchoAndroidView(
     modifier: Modifier = Modifier,
-    provider: IMapViewProvider<T>,
+    factory: (Context) -> T,
+    update: (T, List<Event>, (Event) -> Unit) -> Unit,
+    events: MutableState<List<Event>>,
+    callback: (Event) -> Unit = {}
 ) {
     AndroidView(
+        modifier = modifier,
+        factory = factory,
+        update = { update(it, events.value, callback) }
+    )
+}
+
+/**
+ * This composable will draw a map that displays events.
+ *
+ * @param context Needed to initialize the [MapViewProvider][IMapViewProvider]. Can be the activity
+ *   itself if this function is called from an activity, or the `current` field of a
+ *   [LocalContext][androidx.compose.ui.platform.LocalContext] if it is called from another
+ *   composable.
+ * @param events The list of events to be displayed by the map. Whenever this list changes, the
+ *   event markers in the map will be automatically updated to match the list. Note that this
+ *   parameter is a [MutableState] containing a `List`, so the correct way to instantiate it is the
+ *   following: `val myEvents = remember { mutableStateOf(emptyList<Event>()) }`. Do NOT use the
+ *   `by` keyword. You can then pass it as an argument to this composable. If you want to change the
+ *   value stored in the [MutableState] to trigger a redraw of the markers, you can write
+ *   `myEvents.value = newEventList`.
+ * @param callback The function to be called when a marker on the map is clicked on. The [Event]
+ *   corresponding to the marker will be passed as an argument to this function.
+ */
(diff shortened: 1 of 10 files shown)


alejandrocalles:
This should also close #40. This has been done for `OsmdroidMapViewProvider`. In the case of the `MapLibreMapViewProvider`, the marker features that are shown in most online tutorials/guides (and even in the official `MapLibre` walkthrough) seem to be deprecated. I'm struggling to find appropriate documentation for any other ways to display markers, if there are any.

violoncelloCH:
from a quick search how to draw on the map, I found this: https://maplibre.org/maplibre-native/docs/book/android/annotation-guide.html
is this what you mean seems deprecate? where dd you find that it'd be deprecated?

alejandrocalles:
Yes, that's exactly it. It's tagged as deprecated in the source code. Specifically the function `map.addMarker` and the class `MarkerOptions`. I vaguely remember seeing somewhere that there was another way to do it, but I haven't had much time to look into that.

alejandrocalles:
I did some more research, and the project seems to be in a weird place right now. As you can see [here](https://maplibre.org/maplibre-native/android/api/-map-libre%20-native%20for%20-android/com.mapbox.mapboxsdk.annotations/index.html), the whole `annotations` package has been deprecated.

If you click on any of the classes in the package, they suggest using the [Mapbox annotation plugin](https://github.com/mapbox/mapbox-plugins-android/tree/master/plugin-annotation), which hasn't been updated in 4 years (unless I'm misinterpreting something), and is in a [repository that was archived last year](https://github.com/mapbox/mapbox-gl-native).

I'll try to get it to work using the deprecated features first, and if that fails I'll try out the plugin.

I'm adding these links for future reference:
* [MapLibre Native android changelog](https://github.com/maplibre/maplibre-native/blob/main/platform/android/CHANGELOG.md)
* [Commit where the `annotations` package was deprecated](https://github.com/mapbox/mapbox-gl-native/pull/13613)

app/src/main/java/com/github/swent/echo/compose/map/Map.kt
```diff
@@ -1,59 +1,62 @@
...
+ * @param callback The function to be called when a marker on the map is clicked on. The [Event]
+ *   corresponding to the marker will be passed as an argument to this function.
+ */
 @Composable
 fun MapDrawer(
     modifier: Modifier = Modifier,
-    mapViewFactory: (Context) -> MapView = { createMapView(it, TileSourceFactory.MAPNIK) },
-    update: (MapView) -> Unit = { updateMapView(it, LAUSANNE_GEO_POINT) }
+    context: Context,
+    events: MutableState<List<Event>>,
```
david-schroeter, line 48: I saw that the `events` param is expected to be a `MutableState<...>`. I was wondering why not simply take a `List<Event>` here? So that outside of the `MapDrawer`, it is possible to instantiate the list as `val myEvents by remember { mutableStateOf(emptyList<Event>()) }`. And then simply pass the `myEvents` to the composable, it would prevent the need to call the `.value` when updating the list outside.

I haven't done the test myself, would it prevent the `MapDrawer` from recomposing ?

david-schroeter (commented):
Very good implementation, the integration of the `Provider` abstraction is very well implemented, nice job!

I just added a couple of comments, but in general I found that the code was of great quality! 👍🏻

app/src/main/java/com/github/swent/echo/compose/map/OsmdroidMapViewProvider.kt
```diff
@@ -0,0 +1,87 @@
...
+                // setOnClickListener { ... }
+                controller.setZoom(ZOOM_DEFAULT)
+                controller.setCenter(LAUSANNE_GEO_POINT)
+            }
+        return mapView
+    }
+
+    override fun update(view: MapView, events: List<Event>, callback: (Event) -> Unit) {
+        mapView.apply { drawAllMarkers(events, callback) }
+    }
```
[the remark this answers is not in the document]
alejandrocalles: Thanks for catching this! In principle, the `AndroidView` is responsible for passing the `View` instance to the `update` function, so it's the parameter that should be used in this case.

In fact, the very existence of the `mapView` field is questionable, but I haven't found a better way to properly test my code...

app/src/main/java/com/github/swent/echo/compose/map/Map.kt
alejandrocalles: No, you''re right. I had my doubts about whether the `AndroidView` would be properly recomposed without the `MutableState`, but it seems to work just fine. Thanks!

ntissieres (approved):
Nice work, I tried the fonctionality on your branch and everything works fine. Map is correctly displayed and the pins of the events appears at the good location.

PR #41, opened 2024-04-06
Feature/search tag

srsingh04:
This is the code for the UI of the tags. The tags will be rather personalized for each users depending on what activities they do.

### DIFF
--- a/app/src/androidTest/java/com/github/swent/echo/compose/components/MainScreenTagsTest.kt
+++ b/app/src/androidTest/java/com/github/swent/echo/compose/components/MainScreenTagsTest.kt
@@ -1,0 +1,40 @@
+package com.github.swent.echo.compose.components
+
+import androidx.compose.ui.test.junit4.createComposeRule
+import androidx.compose.ui.test.onNodeWithText
+import androidx.compose.ui.test.performClick
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import com.github.swent.echo.data.model.Tag
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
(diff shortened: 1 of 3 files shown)


ntissieres:
I think we need to discuss more in details how we will implement the tag hierarchy, and follow it for the implementation of this feature. Like if there is more than a number of tags from the same category, we could group them and display the parent tag. When the user click on this parent tag, the children would appear. I will clarify it with you tomorrow if it's not clear.

ntissieres (changes requested):
Maybe it's not a good idea to name the file Tags.kt, we already have a file named Tag to define the data class and moreover it's not truly indicative of what it encodes. MainscreenTagsFilter would be better for example

ntissieres (changes requested):
Don't forget to remove everything that is commented. In order to merge the code into the main, the code should be clean

app/src/main/java/com/github/swent/echo/data/model/Tag.kt
```diff
@@ -1,3 +1,10 @@
...
 
-data class Tag(val tagId: String, val name: String)
+/**
+ * Represents a tag.
+ *
+ * @param tagId The ID of the tag.
+ * @param name The name of the tag.
+ * @param isSelected Whether the tag is selected.
+ */
+data class Tag(val tagId: String, val name: String, val isSelected: Boolean)
```
violoncelloCH, line 10: I'm not sure if having `isSelected` on the data model makes sense. This is state only concerning the UI and nothing we want to store. See also my comment here: https://github.com/echo-swent/echo/pull/31#issuecomment-2047914993

ntissieres (changes requested):
It's better to remove the padding on the top of the items, it will be done after when we add the search tag to the scaffold

ntissieres (approved):
Perfect !

PR #45, opened 2024-04-10
Feature/homescreen-scaffold

### LINKED ISSUE #46: homescreen scaffold
create the scaffold for the homescreen

Greenade:
Created the homescreen composable, as well as the skeleton for the search menu bottom sheet, and the floating button for the search menu. Closes #46


app/src/androidTest/java/com/github/swent/echo/compose/components/TopBarTest.kt
```diff
@@ -1,104 +1,3 @@
...
-    @Test
-    fun shouldCloseHamburgerMenuWhenNavigationItemClicked() {
-        composeTestRule.onNodeWithTag("menu_button").performClick()
-        for (i in 0..6) {
-            composeTestRule.onNodeWithTag("navigation_item_$i").performClick()
-            composeTestRule.onNodeWithTag("top_bar").assertExists()
-        }
-    }
-}
+class TopBarTest {}
```
alejandrocalles, line 3: It might be a good idea to completely remove this file. If we need it in the future, it'll always be possible to find it in the git history.

app/src/main/java/com/github/swent/echo/compose/components/TopBar.kt
```diff
@@ -1,273 +1,11 @@
...
-                        contentDescription = "Search icon to access the search screen"
-                    )
-                }
-            },
-            scrollBehavior = scrollBehavior,
-            modifier = Modifier.testTag("top_bar")
-        )
-    }
-}
+@Composable fun TopBar(navActions: NavigationActions? = null) {}
```
alejandrocalles, line 11: It could also be best to remove this file. What do you think?

app/src/androidTest/java/com/github/swent/echo/compose/components/HomeScreenTest.kt
```diff
@@ -0,0 +1,153 @@
...
+                title = "Bowling Event",
+                description = "",
+                location = Location("Location 1", MAP_CENTER.toGeoPoint()),
+                startDate = Date.from(Instant.now()),
+                endDate = Date.from(Instant.now()),
+                tags = emptySet(),
+                creatorId = "d"
+            )
+        val obj = DisplayEventInfo(event, 0, 0, 0, "")
+    }
```
octogradiste, line 82: We added this test originally to increase the line coverage, but as this doesn't test anything (and we have enough line coverage) we should remove it.

app/src/main/java/com/github/swent/echo/compose/components/HomeScreen.kt
```diff
@@ -0,0 +1,426 @@
...
+                            Icon(
+                                imageVector = Icons.Filled.List,
+                                contentDescription = "Search icon to access the search screen"
+                            )
+                        }
+                    },
+                    scrollBehavior = scrollBehavior,
+                    modifier = Modifier.testTag("top_bar")
+                )
+            },
```
octogradiste, line 316: Can't we extract this into the TopBar.kt file ?

alejandrocalles (dismissed):
A long PR, but a very clean one as well. Thanks a lot for all your work! I left a couple of comments regarding some empty files, but everything else is great :)

octogradiste (commented):
Thanks so much for the hard work!
I just added some minor things which could be changed.
Also, it'd be great if we could do a rebase and clean up the git history. But, this depends on the time you have...

app/src/androidTest/java/com/github/swent/echo/compose/components/TopBarTest.kt
ntissieres: yes i think we can remove it

app/src/main/java/com/github/swent/echo/compose/components/TopBar.kt
ntissieres: also true

srsingh04 (approved):
I think it's ready to merge? There is everything in there no?

Greenade:
> I think it's ready to merge? There is everything in there no?

We still need to delete old files and add the homescreen to the navigation I think

app/src/androidTest/java/com/github/swent/echo/compose/components/HomeScreenTest.kt
violoncelloCH: I've removed that while rebasing and cleaning up the commit history

app/src/main/java/com/github/swent/echo/compose/components/HomeScreen.kt
violoncelloCH: out of scope for now; askip @ntissieres tried to do this but it somehow didn't work... 
I'll create an issue for this...

octogradiste (approved):
LGTM, thanks so much for cleaning up the git history

violoncelloCH:
> LGTM, thanks so much for cleaning up the git history

you're welcome :)

Greenade:
good work everyone :)

PR #49, opened 2024-04-10
Event creation: composables

unglazedstamp:
Add the composables used in the event creation screen.
based on #86 (merge it first)


app/src/main/java/com/github/swent/echo/compose/event/EventTextEntry.kt
```diff
@@ -0,0 +1,66 @@
...
+    value: String,
+    modifier: Modifier = Modifier,
+    onValueChange: (newValue: String) -> Unit
+) {
+    Column(modifier = modifier.padding(EVENT_PADDING_BETWEEN_INPUTS)) {
+        EventEntryName(name = name, modifier = modifier)
+        EventEntryField(
+            value = value,
+            modifier = modifier.testTag("$name-field"),
+            onValueChange = { onValueChange(it) }
```
octogradiste, line 33: Could be just `onValueChange = onValueChange`.

app/src/main/java/com/github/swent/echo/compose/event/EventTagEntry.kt
```diff
@@ -0,0 +1,41 @@
...
+    tagText: String,
+    onTagFieldChanged: (tagText: String) -> Unit,
+    onTagPressed: (tag: Tag) -> Unit
+) {
+    EventTextEntry(name = stringResource(R.string.edit_event_screen_tags), value = tagText) {
+        onTagFieldChanged(it)
+    }
+    Row(modifier = modifier.horizontalScroll(rememberScrollState())) {
+        tags.forEach {
+            TextButton(onClick = { onTagPressed(it) }) {
```
octogradiste, line 31: You could explicitly rename `it` to `tag` to make the code more clear by using `tags.forEach { tag ->`

app/src/main/java/com/github/swent/echo/compose/event/EventScreen.kt
```diff
@@ -0,0 +1,110 @@
...
+            title = title,
+            onBackButtonPressed = { onEventBackButtonPressed() }
+        )
+        // all the inputs for an event
+        EventPropertiesFields(modifier = Modifier, eventViewModel = eventViewModel)
+        // save button
+        OutlinedButton(
+            modifier =
+                Modifier.padding(30.dp).align(Alignment.CenterHorizontally).testTag("Save-button"),
+            onClick = { onEventSaveButtonPressed() }
```
octogradiste, line 44: Could be `onClick = onEventSavedButtonPressed`.

app/src/main/java/com/github/swent/echo/compose/event/EventScreen.kt
```diff
@@ -0,0 +1,110 @@
...
+    }
+}
+
+/**
+ * Modifiable fields of an event: title, description, tags, location, start date, end date, pictures
+ */
+@Composable
+fun EventPropertiesFields(modifier: Modifier, eventViewModel: EventViewModel) {
+    val event by eventViewModel.event.collectAsState()
+    var tagText by remember { mutableStateOf("") }
```
octogradiste, line 57: Could be moved inside the `EventTagEntry` and its `onTagFieldChanged` callback could take as argument the tag text. In this way it would behave more like the `onValueChange` of a `TextField`.

octogradiste (dismissed):
Overall very clean code and pleasant to read! Two minor things I noticed:

- You're passing a `modifier` down to every little composable. I think it's great for modularity, but especially in the case where you use the same `modifier` parameter for two different composables, I doesn't make much sense Also, as those `Event...` composables are used only by you and you're not using this parameter, I think you could drop it in most of the cases. I commented some of them, but there are many more.
- You wrap callbacks: `onClick = { onClick() }` which could be simply `onClick = onClick`.

As always, I'm just nitpicking a lot :)

app/src/main/java/com/github/swent/echo/compose/event/EventTextEntry.kt
```diff
@@ -0,0 +1,66 @@
...
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+
+/** A text entry with a name above */
+@Composable
+fun EventTextEntry(
+    name: String,
+    value: String,
+    modifier: Modifier = Modifier,
```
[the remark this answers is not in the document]
unglazedstamp: It is needed for the test tags

app/src/main/java/com/github/swent/echo/compose/event/EventScreen.kt
unglazedstamp: I let it as it is, because it will be removed by #61 in next sprint

octogradiste (approved):
Thanks for the changes! Looks really clean to me!

unglazedstamp:
Thank you very much for your review. I followed your advice and cleaned the code to improve readability, also removed unnecessary parameters.

PR #52, opened 2024-04-11
Building APK when pushing to main with a version tag

octogradiste:
This is needed for milestone 1. See [deliverables](https://github.com/swent-epfl/public/blob/main/project/handouts/m1-deliverables.md#working-stable-apk).


violoncelloCH (approved):
nice, thanks!

PR #53, opened 2024-04-11
Remove building when pushing on main branch

octogradiste:


violoncelloCH (approved):
LGTM, this way we only build the apk for tags

PR #54, opened 2024-04-11
Feature/hamburger menu

ntissieres:
Implementation of the hamburger menu. I used an already existing composable named ModalNavigationController which appears on the left when we click on the menu button of the top bar, or glide to the right. I created a new data class to display the different items/buttons inside the hamburger menu, in order to create a list and iterate over it to display all the content. The top side of the hamburger menu is a trivial composition of row/column/box.


srsingh04 (dismissed):

Greenade (approved):
nice !

PR #55, opened 2024-04-11
Feature/theme

Greenade:
new themes and new colors


ntissieres (approved):
Be careful to format before commit ! But anyway, it was done after so good job ! Good looking colors ;)

Issue #57, opened 2024-04-12
Find a way to add callbacks on `MapLibre` markers.


alejandrocalles:
Markers should trigger the `callback` function when clicked on.

PR #78, opened 2024-04-15
Add support for callbacks on the `MapLibreMapViewProvider`.

alejandrocalles:
Closing #57.

Add an `onMarkerClickListener` to the `MapLibreMap` when drawing markers.
Change the default provider to be a `MapLibreMapViewProvider`.


octogradiste:
I look into why the marker functions are deprecated. It looks liked there is an effort of splitting the marker logic into a [plugin](https://github.com/mapbox/mapbox-plugins-android/tree/master/plugin-markerview) and to keep the MapLibre core as simple as possible. There are also more [plugins](https://github.com/maplibre/maplibre-plugins-android) available (in case you are interested).

alejandrocalles:
Thanks! I'll look into that.

srsingh04 (approved):
Thank you so much for finding the bug and fixing it with the library versions. 

Good job with the markers implementation overall: maybe, in the future if time permits (could be a feature to implement in the later sprints), I think it could be nice to implement something like marker clustering based on the zoom level and/or the position of the user on the map instead of showing all the markers like the logic does at the moment which may become super messy on the map if there are multiple events to be shown on the map.

alejandrocalles:
Here is the [changelog](https://developer.android.com/jetpack/androidx/releases/compose-runtime#1.6.1) updated version of the `compose` runtime that fixed the bug with the multithread access to a `SnapshotStateObserver`.

In the end I figured that upgrading all the core android/compose libraries would be better.

PR #80, opened 2024-04-15
Modify the `Event` data class.

### LINKED ISSUE #79: Redesign the `Event` data class.
Use `ZonedDateTime` instaed of `java.util.Date`.

Add the fields required by the InfoSheet:
- `organizerName`
- `imageId`
- `participantCount`
- `maxParticipants`

Adapt the rest of the code.

alejandrocalles:
Closes #79.

Change `java.util.Date` to `java.time.ZonedDateTime`. Add fields for `participantsCount`, `maxParticipants`, `organizerName`, `imageId`.

Adapt the rest of the code to match the new `Event` class.

### DIFF
--- a/app/src/main/java/com/github/swent/echo/data/model/Event.kt
+++ b/app/src/main/java/com/github/swent/echo/data/model/Event.kt
@@ -1,6 +1,6 @@
 package com.github.swent.echo.data.model
 
-import java.util.Date
+import java.time.ZonedDateTime
 import kotlinx.serialization.Contextual
 import kotlinx.serialization.Serializable
 
@@ -9,10 +9,14 @@ data class Event(
     val eventId: String,
     val creatorId: String,
     val organizerId: String,
+    val organizerName: String,
     val title: String,
     val description: String,
     val location: Location,
-    @Contextual val startDate: Date,
-    @Contextual val endDate: Date,
-    val tags: Set<Tag>
+    @Contextual val startDate: ZonedDateTime,
+    @Contextual val endDate: ZonedDateTime,
+    val tags: Set<Tag>,
+    val participantCount: Int,
+    val maxParticipants: Int,
+    val imageId: Int
 )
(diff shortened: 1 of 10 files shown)


alejandrocalles:
This is a huge refactor, as it touches a bunch of different files.

It would be ideal if the people who worked on those files could check that the changes are reasonable.

app/src/main/java/com/github/swent/echo/data/model/Event.kt
```diff
@@ -9,10 +9,14 @@ data class Event(
...
     val title: String,
     val description: String,
     val location: Location,
-    @Contextual val startDate: Date,
-    @Contextual val endDate: Date,
-    val tags: Set<Tag>
+    @Contextual val startDate: ZonedDateTime,
+    @Contextual val endDate: ZonedDateTime,
+    val tags: Set<Tag>,
+    val participantCount: Int,
```
Greenade, line 19: not sure about the participantCount, to me it would make more sense that this is linked to a SQL request that views the relationship table between an event and a user (meaning that the user is enrolled in the event) and count on it rather than have a count as a field, for the rest of the changes to event it's good 👍

Greenade (approved):
The changes to HomeScreen and EventInfoSheet seem good to me ! Again I'm not sure about the participantCount field in events, but we could always change it later when the repository is finished. About other files : I have not done them but it looks good to me ! (I may have not noticed tiny details though)

app/src/main/java/com/github/swent/echo/data/model/Event.kt
alejandrocalles: From what I understand, the SQL query to get the participant count would be the responsibility of whoever builds the `Event` object, in this case the repository, but I'm not sure.

@violoncelloCH can you confirm?

violoncelloCH: yes exactly! 
sorry I did not have the time to review this this morning :D

app/src/main/java/com/github/swent/echo/data/model/Event.kt
```diff
@@ -9,10 +9,14 @@ data class Event(
     val eventId: String,
     val creatorId: String,
     val organizerId: String,
+    val organizerName: String,
```
violoncelloCH, line 12: we might prefer to have the objects directly available here rather than the ids and name(s)... (so the organizing association and the creating user(Profile)...

PR #84, opened 2024-04-16
Modify the `Event` and `UserProfile` data classes.

### LINKED ISSUE #79: Redesign the `Event` data class.

alejandrocalles:
Closes #79.

Replace `organizerName` and `organizerId` by field `organizer` of type `Association`.

Replace `creatorId` by field `creator` of type `UserProfile`.

### DIFF
--- a/app/src/main/java/com/github/swent/echo/compose/components/EventInfoSheet.kt
+++ b/app/src/main/java/com/github/swent/echo/compose/components/EventInfoSheet.kt
@@ -88,7 +88,7 @@ fun EventInfoSheet(
             )
             Text(
                 modifier = Modifier.padding(top = 24.dp),
-                text = event.organizerName,
+                text = event.organizer.name,
                 style =
                     TextStyle(
                         fontSize = 20.sp,
--- a/app/src/main/java/com/github/swent/echo/compose/components/EventInfoSheet.kt
+++ b/app/src/main/java/com/github/swent/echo/compose/components/EventInfoSheet.kt
@@ -88,7 +88,7 @@ fun EventInfoSheet(
             )
             Text(
                 modifier = Modifier.padding(top = 24.dp),
-                text = event.organizer.name,
+                text = event.organizer!!.name,
                 style =
                     TextStyle(
                         fontSize = 20.sp,
--- a/app/src/main/java/com/github/swent/echo/data/SampleEvents.kt
+++ b/app/src/main/java/com/github/swent/echo/data/SampleEvents.kt
@@ -2,6 +2,8 @@ package com.github.swent.echo.data
 
 import com.github.swent.echo.compose.map.MAP_CENTER
 import com.github.swent.echo.data.model.Association
+import com.github.swent.echo.data.model.BachelorSection
+import com.github.swent.echo.data.model.BachelorSemester
 import com.github.swent.echo.data.model.Event
 import com.github.swent.echo.data.model.Location
 import com.github.swent.echo.data.model.UserProfile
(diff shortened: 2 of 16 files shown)


app/src/main/java/com/github/swent/echo/data/model/Event.kt
```diff
@@ -7,9 +7,8 @@ import kotlinx.serialization.Serializable
 @Serializable
 data class Event(
     val eventId: String,
-    val creatorId: String,
-    val organizerId: String,
-    val organizerName: String,
+    val creator: UserProfile,
+    val organizer: Association,
```
octogradiste, line 11: We should make this nullable as not every event will be hosted by an association...

octogradiste (changes requested):
I like that you introduced a list of sample events!

alejandrocalles:
Thanks for the review! The list of sample events should be temporary, but I hope it will make refactoring easier.

app/src/main/java/com/github/swent/echo/compose/components/EventInfoSheet.kt
```diff
@@ -90,7 +90,7 @@ fun EventInfoSheet(
             )
             Text(
                 modifier = Modifier.padding(top = 24.dp),
-                text = event.organizerName,
+                text = event.organizer!!.name,
```
violoncelloCH, line 93: I think this needs a safety check somewhere to not crash the app if organizer is null... the organizer should just not be showed (but this might be out of scope for this PR)

app/src/main/java/com/github/swent/echo/data/SampleEvents.kt
```diff
@@ -0,0 +1,64 @@
...
+import com.github.swent.echo.data.model.BachelorSection
+import com.github.swent.echo.data.model.BachelorSemester
+import com.github.swent.echo.data.model.Event
+import com.github.swent.echo.data.model.Location
+import com.github.swent.echo.data.model.UserProfile
+import java.time.ZonedDateTime
+
+// This should come from the repository
+
+val SAMPLE_EVENTS: List<Event> =
```
violoncelloCH, line 14: I guess these `SAMPLE_EVENTS` is just temporary until the repository is able to return events. Maybe create an issue so we don't forget to adapt and remove these samples.

app/src/main/java/com/github/swent/echo/data/model/EPFLSection.kt
```diff
@@ -0,0 +1,30 @@
+package com.github.swent.echo.data.model
+
+interface EPFLSection
+
+enum class BachelorSection : EPFLSection {
```
violoncelloCH, line 5: I think it would make sense to rather have an interface named just `Section` and have the class named something like `EPFLBcheorSection` etc. This way the interface is generic...

app/src/main/java/com/github/swent/echo/data/model/EPFLSemester.kt
```diff
@@ -0,0 +1,20 @@
+package com.github.swent.echo.data.model
+
+interface EPFLSemester
+
+enum class BachelorSemester : EPFLSemester {
```
violoncelloCH, line 5: same for the naming as for the Sections class ^

violoncelloCH (approved):
thanks for working on this!
looks good otherwise!

app/src/main/java/com/github/swent/echo/compose/components/EventInfoSheet.kt
alejandrocalles: Added this to a new issue: #98 .

app/src/main/java/com/github/swent/echo/data/SampleEvents.kt
alejandrocalles: Added this to a new issue: #100.

violoncelloCH (approved):
thanks for implementing the changes I requested and opening the issues for things to tackle later :)
looks all good now!

PR #82, opened 2024-04-16
Display authentication error messages

octogradiste:
This will close #81. The goal is to give to the user more feedback on why the authentication process fails. 
I transform the error messages from Supabase to a human-readable format and display them to the user.

PR #85, opened 2024-04-16
Stay logged in over multiple app restarts

### LINKED ISSUE #83: As a logged in user, I want to be redirected directly to the map screen when I start the app, so I don't have to login every time.

octogradiste:
Users won't have to log in every time they start the app.

This requires #82 to be merged first and will close #83.

### DIFF
--- a/app/src/main/java/com/github/swent/echo/MainActivity.kt
+++ b/app/src/main/java/com/github/swent/echo/MainActivity.kt
@@ -7,14 +7,24 @@ import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Surface
 import androidx.compose.ui.Modifier
+import com.github.swent.echo.authentication.AuthenticationService
 import com.github.swent.echo.compose.navigation.AppNavigationHost
 import com.github.swent.echo.ui.theme.EchoTheme
 import dagger.hilt.android.AndroidEntryPoint
+import javax.inject.Inject
+import kotlinx.coroutines.runBlocking
 
 @AndroidEntryPoint
 class MainActivity : ComponentActivity() {
+
+    @Inject lateinit var authenticationService: AuthenticationService
+
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
+
+        // This happens so quickly, I doesn't make sense to show a loading screen.
+        runBlocking { authenticationService.initialize() }
+
         setContent {
             EchoTheme {
                 // A surface container using the 'background' color from the theme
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/github/swent/echo/MainActivity.kt
```diff
@@ -7,22 +7,32 @@ import androidx.compose.foundation.layout.fillMaxSize
...
 @AndroidEntryPoint
 class MainActivity : ComponentActivity() {
+
+    @Inject lateinit var authenticationService: AuthenticationService
+
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
+
+        // This happens so quickly, I doesn't make sense to show a loading screen.
+        runBlocking { authenticationService.initialize() }
```
unglazedstamp, line 26: This should at least be in the AppNavigationHost, because the authenticationService isn't necessary in the MainActivity

app/src/main/java/com/github/swent/echo/authentication/AuthenticationService.kt
```diff
@@ -6,6 +6,9 @@ import io.github.jan.supabase.compose.auth.composable.NativeSignInResult
 /** Service to handle user authentication. */
 interface AuthenticationService {
 
+    /** Initialize the authentication service. */
+    suspend fun initialize()
```
unglazedstamp, line 10: Why is it needed to expose an initialize function? It may be simpler to add an init block in the implementation of the authentication service.

app/src/main/java/com/github/swent/echo/authentication/AuthenticationService.kt
```diff
@@ -36,7 +39,19 @@ interface AuthenticationService {
...
      */
-    suspend fun getCurrentUserID(): String?
+    fun getCurrentUserID(): String?
+
+    /**
+     * Check if the user is logged in.
+     *
+     * @return True if the user is logged in, false otherwise.
+     */
+    fun userIsLoggedIn(): Boolean {
```
unglazedstamp, line 54: Then this could be a stateflow to just recompose the AppNavigationHost when the user is logged in

unglazedstamp (changes requested):
Good feature, However It feels like exposing an initialize function for the AuthenticationService isn't necessary.

app/src/main/java/com/github/swent/echo/MainActivity.kt
octogradiste: The `AppNavigationHost` is a composable, every time it recomposes it would reinitialize the `AuthenticationService`.

octogradiste:
Thanks so much for the review! As we discussed earlier, we will extract the initialization logic into a separate loading screen. I opened an issue #105 for this!

unglazedstamp (approved):
All good, thank you !

PR #86, opened 2024-04-17
Integration of the repository in the event viewmodel

### LINKED ISSUE #60: Add the reopsitory to the event viewmodel

unglazedstamp:
Use hilt to inject a repository and an authentication service in the viewmodel.
Adapt the getters and setters to use the repository.
Will close #60

### DIFF
--- a/app/src/main/java/com/github/swent/echo/viewmodels/event/EventViewModel.kt
+++ b/app/src/main/java/com/github/swent/echo/viewmodels/event/EventViewModel.kt
@@ -3,6 +3,7 @@ package com.github.swent.echo.viewmodels.event
 import android.util.Log
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
+import com.github.swent.echo.R
 import com.github.swent.echo.authentication.AuthenticationService
 import com.github.swent.echo.data.model.Association
 import com.github.swent.echo.data.model.Event
@@ -49,7 +50,8 @@ constructor(
         viewModelScope.launch {
             val userid = authenticationService.getCurrentUserID()
             if (userid == null) {
-                _status.value = EventStatus.Error("you are not logged in")
+                _status.value = EventStatus.Error(R.string.event_creation_error_not_logged_in)
+                Log.e("create event", "the user is not logged in")
             } else {
                 _event.value =
                     _event.value.copy(
@@ -176,10 +178,10 @@ constructor(
         val event = _event.value
         if (event.startDate.isAfter(event.endDate)) {
             _status.value =
-                EventStatus.Error("end date before start date") // TODO: use error code from R
+                EventStatus.Error(R.string.event_creation_error_end_date_before_start_date)
         }
         if (event.title.isBlank()) {
-            _status.value = EventStatus.Error("title is empty")
+            _status.value = EventStatus.Error(R.string.event_creation_error_empty_title)
         }
         return _status.value !is EventStatus.Error
     }
@@ -196,5 +198,5 @@ sealed class EventStatus {
     // syncing with the version in the repository
     data object Saving : EventStatus()
 
-    data class Error(val error: String) : EventStatus()
+    data class Error(val errorRef: Int) : EventStatus()
 }
--- a/app/src/main/java/com/github/swent/echo/viewmodels/event/EventViewModel.kt
+++ b/app/src/main/java/com/github/swent/echo/viewmodels/event/EventViewModel.kt
@@ -6,18 +6,12 @@ import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import com.github.swent.echo.R
 import com.github.swent.echo.authentication.AuthenticationService
-import com.github.swent.echo.data.model.Association
 import com.github.swent.echo.data.model.Event
-import com.github.swent.echo.data.model.Location
 import com.github.swent.echo.data.model.Tag
-import com.github.swent.echo.data.model.UserProfile
 import com.github.swent.echo.data.repository.Repository
 import dagger.hilt.android.lifecycle.HiltViewModel
-import java.time.ZonedDateTime
-import com.github.swent.echo.data.repository.Repository
 import javax.inject.Inject
 import kotlinx.coroutines.flow.MutableStateFlow
-import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.launch
 
@@ -38,6 +32,8 @@ constructor(
     private val allTagsList = MutableStateFlow<List<Tag>>(listOf())
     private val _isEventNew = MutableStateFlow(true)
     val isEventNew = _isEventNew.asStateFlow()
+    private val _organizerList = MutableStateFlow<List<String>>(listOf())
+    val organizerList = _organizerList.asStateFlow()
 
     // initialize async values
     init {
@@ -48,6 +44,7 @@ constructor(
                 _status.value = EventStatus.Error(R.string.event_creation_error_not_logged_in)
                 Log.e("create event", "the user is not logged in")
             } else {
+                val username = repository.getUserProfile(userid).name
                 if (savedEventId.contains("eventId")) {
                     _isEventNew.value = false
                     _event.value = repository.getEvent(savedEventId.get<String>("eventId")!!)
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/github/swent/echo/viewmodels/event/EventViewModel.kt
```diff
@@ -29,30 +43,77 @@ class EventViewModel() : ViewModel() {
...
+            _status.value = EventStatus.Saved
+        }
     }
 
-    // return the organizer name of the event
-    fun getOrganizerName(): String {
-        // TODO: get organizer name from repository
-        return ""
+    // return the event
+    fun getEvent(): StateFlow<Event> {
```
octogradiste, line 83: This could be simply a field, i.e. `val event = _event.asStateFlow()`. Then in the composable you will be able to use the value of the `StateFlow` to automatically trigger the recomposition. Otherwise, if you simply use the `Event` in you composable, it won't automatically trigger the recomposition...

app/src/main/java/com/github/swent/echo/viewmodels/event/EventViewModel.kt
```diff
@@ -93,25 +154,35 @@ class EventViewModel() : ViewModel() {
...
+                    _status.value = EventStatus.Saved
+                }
             }
         }
     }
 
     // return the status of the event
-    fun getStatus(): EventStatus {
-        return _status.asStateFlow().value
+    fun getStatus(): StateFlow<EventStatus> {
```
octogradiste, line 173: This should be a simple field `val status = _status.asStateFlow()`

app/src/main/java/com/github/swent/echo/viewmodels/event/EventViewModel.kt
```diff
@@ -29,30 +43,77 @@ class EventViewModel() : ViewModel() {
         )
     private val _event = MutableStateFlow<Event>(emptyEvent)
```
octogradiste, line 44: Nitpicking: It would be very clean if we could use a constant like `Event.EMPTY`. We would need to add a `companion object { const val EMPTY = Event(...) }` to the `Event` data class.

app/src/main/java/com/github/swent/echo/viewmodels/event/EventViewModel.kt
```diff
@@ -29,30 +43,77 @@ class EventViewModel() : ViewModel() {
...
+    constructor(
+        repository: Repository,
+        authenticationService: AuthenticationService,
+        eventId: String
+    ) : this(repository, authenticationService) {
+        _status.value = EventStatus.Saving // avoid saving before initialization
+        viewModelScope.launch {
+            _event.value = repository.getEvent(eventId)
+            _status.value = EventStatus.Saved
+        }
```
octogradiste, line 79: Unsure how this is going to work. Normally hilt will automatically create the view model and inject the dependencies... When will this constructor be called ? I think the solution is to use see [ViewModel SavedStateHandle](https://developer.android.com/topic/libraries/architecture/viewmodel/viewmodel-savedstate) and [Hilt SavedStateHandle](https://developer.android.com/training/dependency-injection/hilt-jetpack). The `SavedStateHandle` is like a map, you could try to get the `eventId` from it. If it returns `null` you need to create a new event otherwise you modify the event with the given `eventId`. Also, in this case you don't need to get the user id as it will already be saved in the `Event`.

app/src/main/java/com/github/swent/echo/viewmodels/event/EventViewModel.kt
```diff
@@ -29,30 +43,77 @@ class EventViewModel() : ViewModel() {
         )
     private val _event = MutableStateFlow<Event>(emptyEvent)
     private val _status = MutableStateFlow<EventStatus>(EventStatus.New)
+    private val allTagsList = MutableStateFlow<List<Tag>>(listOf())
 
-    // placeholder for the list of all tags
-    private val allTagsList = listOf(Tag("1", "tag1"), Tag("2", "tag2"), Tag("3", "tag3"))
+    // initialize async values
+    init {
+        _status.value = EventStatus.Saving // avoid saving before initialization
```
octogradiste, line 50: Nitpicking: Maybe you could add an `EventStatus.Initializing`, might be more clear...

app/src/main/java/com/github/swent/echo/viewmodels/event/EventViewModel.kt
```diff
@@ -29,30 +43,77 @@ class EventViewModel() : ViewModel() {
...
+    // return the event
+    fun getEvent(): StateFlow<Event> {
+        return _event.asStateFlow()
     }
 
     // return the list of possible organizer for the user
-    fun getOrganizerList(): List<String> {
-        // placeholder until we get organizer list from repository
-        return listOf("testOrganizer", "anotherTestOrganizer")
+    fun getOrganizerList(): StateFlow<List<String>> {
```
octogradiste, line 88: This shuold be a simple filed `val organizers = _organizers.asStateFlow()`.

app/src/main/java/com/github/swent/echo/viewmodels/event/EventViewModel.kt
```diff
@@ -29,30 +43,77 @@ class EventViewModel() : ViewModel() {
...
+                }
+            }
+            if (organizerId != "") {
+                setEvent(
+                    _event.value.copy(organizerId = organizerId, organizerName = organizerName)
+                )
+            } else {
+                Log.e("set event organizer", "organizer not found in the repository")
+            }
+        }
```
octogradiste, line 116: Do we need to do all those checks as the user can only pick from the organizers list you show to him. If we only show him the organizations he's in, he won't be able to chose an incompatible association. In the worst case (he gets kicked out of the association during the creation phase) the security rules of supabase will prevent saving to the database.

Maybe this function is not even necessary as the composable will simply call `setEvent` with the organizer field already set according to the users actions.

octogradiste (changes requested):
Nice work! There are just some things to change, to make it easier to use in a composable.

app/src/main/java/com/github/swent/echo/viewmodels/event/EventViewModel.kt
unglazedstamp: Thanks for the reasoning, I changed this function to set only the creator as organizer and added a todo until we get #84 merged to add association as organizer

app/src/main/java/com/github/swent/echo/viewmodels/event/EventViewModel.kt
```diff
@@ -1,59 +1,81 @@
...
+    private val savedEventId: SavedStateHandle
+) : ViewModel() {
 
-    // placeholder for the list of all tags
-    private val allTagsList = listOf(Tag("1", "tag1"), Tag("2", "tag2"), Tag("3", "tag3"))
+    private val _event = MutableStateFlow<Event>(Event.EMPTY)
+    private val _status = MutableStateFlow<EventStatus>(EventStatus.Modified)
+    val event = _event.asStateFlow()
+    val status = _status.asStateFlow()
+    private val allTagsList = MutableStateFlow<List<Tag>>(listOf())
```
octogradiste, line 32: Does this needs to be a `MutableStateFlow` as it won't be exposed in any composable? This could simply be a list of tags (`private lateinit var allTags: List<Tag>`) which would be initialized in the `init` block.

app/src/main/java/com/github/swent/echo/viewmodels/event/EventViewModel.kt
```diff
@@ -95,39 +115,42 @@ class EventViewModel() : ViewModel() {
...
 /** the different status of an event */
 sealed class EventStatus {
-    // new event not in the repository
-    data object New : EventStatus()
     // same as in the repository
     data object Saved : EventStatus()
     // different from the version in the repository
     data object Modified : EventStatus()
     // syncing with the version in the repository
     data object Saving : EventStatus()
```
octogradiste, line 153: Nitpicking: maybe `Syncing` would be more clear

app/src/main/java/com/github/swent/echo/viewmodels/event/EventViewModel.kt
```diff
@@ -1,59 +1,81 @@
...
+constructor(
+    private val repository: Repository,
+    private val authenticationService: AuthenticationService,
+    private val savedEventId: SavedStateHandle
+) : ViewModel() {
 
-    // placeholder for the list of all tags
-    private val allTagsList = listOf(Tag("1", "tag1"), Tag("2", "tag2"), Tag("3", "tag3"))
+    private val _event = MutableStateFlow<Event>(Event.EMPTY)
+    private val _status = MutableStateFlow<EventStatus>(EventStatus.Modified)
```
octogradiste, line 29: Could be `EventStatus.Saving` so you can remove `_event.value = EventStatus.Saving` from the `init` block.

octogradiste (changes requested):

app/src/main/java/com/github/swent/echo/viewmodels/event/EventViewModel.kt
unglazedstamp: I prefer to keep `Saving` for now, because the button in the UI is called "Save". Maybe when add the edit event screen we could rename it.

octogradiste (approved):
Thanks! LGTM

Issue #51, opened 2024-04-11
Create Tables with RLS on Supabase


violoncelloCH:
Create tables on supabase with corresponding RLS (Row Level Security)

- [x] create tables
- [x] configure RLS: https://supabase.com/docs/guides/database/postgres/row-level-security

violoncelloCH:
okay, so I did add the tables and the RLS policies to on Supabase.
@octogradiste could you review especially the policies? Thanks

violoncelloCH:
Ohh indeed, you're right!
Though that's actually just a selector for which role the rule applies to... public just means all roles, so I suppose it doesn't really make a difference, but I'm not completely sure. Would need to investigate further, but for now I changed the rule to only apply to authenticated users to be sure.
Thanks for the review!

PR #87, opened 2024-04-17
👌 IMPROVE: update Supabase DB table names and attributes to snake_cas…

violoncelloCH:
…e to match Supabase/Postgres requirements

Postgres policies for row level security (RLS) require the table names and attributes to be in snake_case rather than camelCase. I therefore changed this in the Supabase RemoteDataSource implementation as well.

in accordance with #51


violoncelloCH:
@octogradiste should be all ready now... SonarCloud is complaining obviously since we disabled the tests for the SupabaseDataSource

octogradiste (approved):
Thanks so much!! This will fix the pipeline so we can continue pushing to main. :)
There is still some work needed to handle the location of an event, but this some later PR.

violoncelloCH:
Indeed, I'm already working on correctly fetching an `Event`; it's not trivial to make the thing de-serialize correctly :/

PR #91, opened 2024-04-18
Feature/search tool

ntissieres:
I implemented here the search tool in order to apply filters on the events to display. HomeScreen have now a FilterContainer to store all differents filters value, that it passes to the SearchMenuSheet and then to the SearchMenuFilters.  SearchMenuSheet is the content of the bottom sheet who display either the filter mode, either the discover mode. Their will be respectively implemented in the SearchMenuFilter and SearchMenuDiscover.


ntissieres:
Still need to adapt the test and coverage...

PR #106, opened 2024-04-19
Adding  content inside the search menu

ntissieres:
First of all, i created a dataclass FiltersContainer which contains every filters we will use for the events to display. HomeScreen passes as argument to SearchMenuSheet a FiltersContainer to get the different filters. SearchMenuSheet is the component that display the search menu, and it is in charge of completing the tagname/id and can switch between filters mode or discover mode. According to the user, SearchMenuFilters or SearchMenuDiscover is displayed. SearchMenuFilters contains checkboxes to filter between epfl, section, class, pending, confirmed or full. By the end of the next sprint, it will also allowed to choose a period of time for the event. SearchMenuDiscover is actually not implemented.


app/src/main/java/com/github/swent/echo/compose/components/searchmenu/FiltersContainer.kt
```diff
@@ -0,0 +1,27 @@
...
+
+data class FiltersContainer(
+    var tagId: MutableState<String>,
+    var epflChecked: MutableState<Boolean>,
+    var sectionChecked: MutableState<Boolean>,
+    var classChecked: MutableState<Boolean>,
+    var pendingChecked: MutableState<Boolean>,
+    var confirmedChecked: MutableState<Boolean>,
+    var fullChecked: MutableState<Boolean>,
+    @Contextual var from: MutableState<ZonedDateTime>,
```
unglazedstamp, line 15: Why @Contextual is used here ? FiltersContainers has to be serializable ?

app/src/main/java/com/github/swent/echo/compose/components/searchmenu/SearchMenuFilters.kt
```diff
@@ -0,0 +1,198 @@
...
+        }
+    }
+}
+
+/*
+ * Composable to get a date from the user.
+ * TODO : Implement it, but I spend a bit too much time
+ *   on non-working solution so leaving it for next sprint
+ */
+@Composable fun DateInputSample(dateOutput: MutableState<ZonedDateTime>) {}
```
violoncelloCH, line 198: Hi, for the time selector slider, this component looks like it does what we need: https://developer.android.com/develop/ui/compose/components/slider
Did you try this?

unglazedstamp (dismissed):
Thanks a lot for these improvements !

Greenade (approved):
nice work ! love the look of the search sheet ! the force push to make a clean PR was a good move... 👍

PR #109, opened 2024-04-19
event and userProfile: fetch/set from supabase with all relations

### LINKED ISSUE #68: [repository] implement tags and participant relation for event and userProfile and fetch them correctly from supbase
implement the following in the form of relations:
- [x] tags on events
- [x] participants on events
- [x] tags on user_profiles

violoncelloCH:
correctly serialize and translate nested objects to DB relations (and inversely)
closes #68

### DIFF
--- a/app/src/main/java/com/github/swent/echo/data/supabase/SupabaseDataSource.kt
+++ b/app/src/main/java/com/github/swent/echo/data/supabase/SupabaseDataSource.kt
@@ -15,9 +15,7 @@ import io.github.jan.supabase.SupabaseClient
 import io.github.jan.supabase.postgrest.from
 import io.github.jan.supabase.postgrest.query.Columns
 
-class SupabaseDataSource(supabaseClient: SupabaseClient) : RemoteDataSource {
-
-    var supabase = supabaseClient
+class SupabaseDataSource(private val supabase: SupabaseClient) : RemoteDataSource {
 
     override suspend fun getAssociation(associationId: String): Association {
         return supabase
@@ -30,17 +28,16 @@ class SupabaseDataSource(supabaseClient: SupabaseClient) : RemoteDataSource {
         return supabase.from("associations").select().decodeList<Association>()
     }
 
+    companion object {
+        const val QUERY_EVENT =
+            "event_id, user_profiles!public_events_creator_id_fkey(user_id, name), associations(association_id, name, description), title, description, event_tags(tags(tag_id, name, parent_id)), location_name, location_lat, location_long, start_date, end_date, participant_count, max_participants, image_id"
+    }
+
     override suspend fun getEvent(eventId: String): Event {
         var event =
             supabase
                 .from("events")
-                .select(
-                    Columns.raw(
-                        "event_id, user_profiles!public_events_creator_id_fkey(user_id, name), associations(association_id, name, description), title, description, event_tags(tags(tag_id, name)), location_name, location_lat, location_long, start_date, end_date, participant_count, max_participants, image_id"
-                    )
-                ) {
-                    filter { eq("event_id", eventId) }
-                }
+                .select(Columns.raw(QUERY_EVENT)) { filter { eq("event_id", eventId) } }
                 .decodeSingle<EventSupabase>()
         return event.toEvent()
     }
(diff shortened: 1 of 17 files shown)


app/src/main/java/com/github/swent/echo/data/supabase/SupabaseDataSource.kt
```diff
@@ -0,0 +1,98 @@
...
+import com.github.swent.echo.data.supabase.entities.EventSupabaseSetter
+import com.github.swent.echo.data.supabase.entities.EventTagSupabase
+import com.github.swent.echo.data.supabase.entities.UserProfileSupabase
+import com.github.swent.echo.data.supabase.entities.UserProfileSupabaseSetter
+import com.github.swent.echo.data.supabase.entities.UserTagSupabase
+import io.github.jan.supabase.SupabaseClient
+import io.github.jan.supabase.postgrest.from
+import io.github.jan.supabase.postgrest.query.Columns
+
+class SupabaseDataSource(supabaseClient: SupabaseClient) : RemoteDataSource {
```
octogradiste, line 18: You could use `(private val supabase: SupabaseClient)` instead and remove `var supabase = supabaseClient`.

app/src/main/java/com/github/swent/echo/data/supabase/SupabaseDataSource.kt
```diff
@@ -0,0 +1,98 @@
...
+        return supabase.from("associations").select().decodeList<Association>()
+    }
+
+    override suspend fun getEvent(eventId: String): Event {
+        var event =
+            supabase
+                .from("events")
+                .select(
+                    Columns.raw(
+                        "event_id, user_profiles!public_events_creator_id_fkey(user_id, name), associations(association_id, name, description), title, description, event_tags(tags(tag_id, name)), location_name, location_lat, location_long, start_date, end_date, participant_count, max_participants, image_id"
```
octogradiste, line 39: Isn't the id of the parent tag missing? Or maybe I can just not read SQL queries...

app/src/main/java/com/github/swent/echo/data/supabase/SupabaseDataSource.kt
```diff
@@ -0,0 +1,98 @@
...
+        supabase.from("event_tags").upsert(eventTags)
+    }
+
+    override suspend fun getAllEvents(): List<Event> {
+        var events =
+            supabase
+                .from("events")
+                .select(
+                    Columns.raw(
+                        "event_id, user_profiles!public_events_creator_id_fkey(user_id, name), associations(association_id, name, description), title, description, event_tags(tags(tag_id, name)), location_name, location_lat, location_long, start_date, end_date, participant_count, max_participants, image_id"
```
octogradiste, line 62: Id of parent tag is probably missing.

app/src/main/java/com/github/swent/echo/data/supabase/SupabaseDataSource.kt
```diff
@@ -0,0 +1,98 @@
...
+    override suspend fun getAllTags(): List<Tag> {
+        return supabase.from("tags").select().decodeList<Tag>()
+    }
+
+    override suspend fun getUserProfile(userId: String): UserProfile {
+        var userProfile =
+            supabase
+                .from("user_profiles")
+                .select(
+                    Columns.raw("user_id, name, semester, section, user_tags(tags(tag_id, name))")
```
octogradiste, line 82: Id of parent tag is probably missing.

octogradiste (changes requested):
Thanks so much!! The code is really easy to read and of good quality :) One thing I'm not sure is if you are actually getting the `parentId` of the tags properly.

app/src/main/java/com/github/swent/echo/data/supabase/SupabaseDataSource.kt
octogradiste: Also, I forgot to add: this should be a constant as you are using it twice. To add constant in Kotlin, you can use `companion object`s:
```kotlin
companion object {
  const val QUERY = "...."
}
```
Just add the `companion object` inside your class and use `QUERY` as constant.

app/src/main/java/com/github/swent/echo/data/supabase/SupabaseDataSource.kt
violoncelloCH: you're right... I just moved this file actually and didn't touch this part o the code, but it makes sense obviously! I'll change that.

app/src/main/java/com/github/swent/echo/data/supabase/SupabaseDataSource.kt
violoncelloCH: You're absolutely right that `parent_tag` field is missing; great catch! This was actually changed while I've already been working on it and I didn't realize it while rebasing. Since the fields is nullable I also didn't catch this during testing. I'll add the `parent_tag`.

And yes, that makes sense to put that query into a constant. Thanks for indicating how to do this :)

octogradiste (approved):
Looks good to me!

PR #112, opened 2024-04-22
Define the interface of the local data source

octogradiste:


app/src/main/java/com/github/swent/echo/data/repository/datasources/LocalDataSource.kt
```diff
@@ -1,5 +1,53 @@
...
+/**
+ * The local data source.
+ *
+ * Note that all the getters return null when the data is not found locally.
+ */
 interface LocalDataSource {
-    fun getData(): String
+
+    /** Initialize the local data source. */
+    suspend fun initialize()
```
violoncelloCH, line 16: I'm wondering if there is any interest in having this here. Anyways initializing the DB should be a one-time action that's never repeated. Does it make sense for the consumer of the LocalDataSource (i.e. the Repository) to care about whether Room is initialized? Would it maybe make more sense to not expose this and just have room initialize itself either on the first startup of the app, a hook during login/registration or simply when one of it's other methods is called for the first time ?

violoncelloCH (dismissed):
Solid interface design. I can't see anything missing for now. Just a small comment or rather question below.

app/src/main/java/com/github/swent/echo/data/repository/datasources/LocalDataSource.kt
octogradiste: You're right! I added this as I thought we might need to do some initialization logic, but after starting implementing the interface and reading through the docs of room, I didn't find anything that should be initialized... Therefore, I removed it.

violoncelloCH: okay great :)

violoncelloCH (approved):

PR #113, opened 2024-04-22
Feature/tag selection composable

unglazedstamp:
The tag selection dialog composable, need the tag viewmodel merged first #111

### DIFF
--- a/app/src/main/java/com/github/swent/echo/compose/components/TagSelectionDialog.kt
+++ b/app/src/main/java/com/github/swent/echo/compose/components/TagSelectionDialog.kt
@@ -40,10 +40,13 @@ fun TagSelectionDialog(
     val currentDepth = tagViewModel.currentDepth.collectAsState()
     val hasSubTags = currentDepth.value < tagViewModel.maxDepth
     Dialog(onDismissRequest = onDismissRequest, properties = dialogProperties) {
-        Card(modifier = Modifier.padding(vertical = 30.dp)) {
+        Card(modifier = Modifier.padding(vertical = 30.dp).testTag("tag-dialog")) {
             Column(modifier = Modifier.verticalScroll(rememberScrollState())) {
                 Row {
-                    IconButton(onClick = tagViewModel::goUp) {
+                    IconButton(
+                        onClick = tagViewModel::goUp,
+                        modifier = Modifier.testTag("tag-back-button")
+                    ) {
                         val icon = Icons.Filled.KeyboardArrowLeft
                         Icon(imageVector = icon, contentDescription = icon.name)
                     }
@@ -77,10 +80,15 @@ fun TagSelectionDialogEntry(
         modifier = Modifier.padding(5.dp).fillMaxWidth(),
         horizontalArrangement = Arrangement.SpaceBetween
     ) {
-        TextButton(onClick = { onTagClicked(tag) }, modifier = Modifier) { Text(tag.name) }
+        TextButton(
+            onClick = { onTagClicked(tag) },
+            modifier = Modifier.testTag("${tag.name}-select-button")
+        ) {
+            Text(tag.name)
+        }
         if (hasSubTags) {
             IconButton(
-                modifier = Modifier.testTag("${tag.name}-button"),
+                modifier = Modifier.testTag("${tag.name}-subtag-button"),
                 onClick = { onTagArrowClicked(tag) }
             ) {
                 val icon = Icons.Filled.KeyboardArrowRight
--- a/app/src/main/java/com/github/swent/echo/compose/components/TagSelectionDialog.kt
+++ b/app/src/main/java/com/github/swent/echo/compose/components/TagSelectionDialog.kt
@@ -1,0 +1,91 @@
+package com.github.swent.echo.compose.components
+
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.rememberScrollState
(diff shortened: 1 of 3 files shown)


ntissieres (changes requested):
While the repository still not functional, I don't think it's a good idea to approve a code that would crash if used. It may be preferable to comment every use of the repository and replace it by an intermediate compilable version of it, in order to test it in the emulator to see if everything looks fine (more importantly for UI). For example, you could comment every call to the repository and put direct value instead, and accept a nullable repository for TagViewModel.

unglazedstamp:
I agree with you. However this is addressed in #136 which replaces the remote database by a local storage, once it's done you will be able to launch the app and check the UI. If I hardcode tags now, I will have to revert these changes later, so I prefer to wait for the local storage to be merged so that we don't have to go back to each hardcoded composable of the project when we connect the remote database.

ntissieres:
Good point. Considering the code, it seems perfect and ready to be merged after the #136

app/src/main/java/com/github/swent/echo/compose/components/TagSelectionDialog.kt
```diff
@@ -0,0 +1,99 @@
...
+fun TagSelectionDialog(
+    onDismissRequest: () -> Unit,
+    dialogProperties: DialogProperties = DialogProperties(),
+    tagViewModel: TagViewModel,
+    onTagSelected: (Tag) -> Unit
+) {
+    val tags = tagViewModel.tags.collectAsState()
+    val currentDepth = tagViewModel.currentDepth.collectAsState()
+    val hasSubTags = currentDepth.value < tagViewModel.maxDepth
+    Dialog(onDismissRequest = onDismissRequest, properties = dialogProperties) {
```
ntissieres, line 42: It would be better to have a fixed size Dialog (modifier.width(...).height(...)), because actually the size of the composable change each time to fit what to display

app/src/main/java/com/github/swent/echo/compose/components/TagSelectionDialog.kt
```diff
@@ -0,0 +1,99 @@
...
+        Card(modifier = Modifier.padding(vertical = 30.dp).testTag("tag-dialog")) {
+            Column(modifier = Modifier.verticalScroll(rememberScrollState())) {
```
ntissieres, line 44: LazyColumn is more adapted than a Column for scrollable list, it is more optimized (https://stackoverflow.com/questions/73734918/android-compose-difference-between-lazycolumn-and-column-with-verticalscroll)

app/src/main/java/com/github/swent/echo/compose/components/TagSelectionDialog.kt
```diff
@@ -0,0 +1,99 @@
...
+                Row {
+                    IconButton(
+                        onClick = tagViewModel::goUp,
+                        modifier = Modifier.testTag("tag-back-button")
+                    ) {
+                        val icon = Icons.Filled.KeyboardArrowLeft
+                        Icon(imageVector = icon, contentDescription = icon.name)
+                    }
+                }
+                Spacer(Modifier.padding(5.dp))
```
ntissieres, line 54: Maybe Modifier.height would make more sense here, padding add 5.dp on top and on bottom so 10.dp and is not really meant for that type of situation

app/src/main/java/com/github/swent/echo/compose/components/TagSelectionDialog.kt
```diff
@@ -0,0 +1,99 @@
...
+                for (tag in tags.value) {
+                    TagSelectionDialogEntry(
```
ntissieres, line 56: I'm not certain it's really intuitive to click on the text to select a tag, I would say we're used to click on the row (text + arrow included) to dive into the subtag, and maybe have an "add" icon after the arrow to effectivelly select the tag

app/src/main/java/com/github/swent/echo/compose/components/TagSelectionDialog.kt
```diff
@@ -0,0 +1,99 @@
...
+        modifier = Modifier.padding(5.dp).fillMaxWidth(),
+        horizontalArrangement = Arrangement.SpaceBetween
+    ) {
+        TextButton(
+            onClick = { onTagClicked(tag) },
+            modifier = Modifier.testTag("${tag.name}-select-button")
+        ) {
+            Text(tag.name)
+        }
+        if (hasSubTags) {
```
ntissieres, line 89: Maybe it's an error on the simple repository but the hasSubTags doesn't work, it still appears when we reach the "childest" tag (like Football) and reach a page with nothing

ntissieres (changes requested):
Overall the tag selection works fine but some little changes still required

app/src/main/java/com/github/swent/echo/compose/components/TagSelectionDialog.kt
unglazedstamp: That's true, I fixed it

app/src/main/java/com/github/swent/echo/compose/components/TagSelectionDialog.kt
unglazedstamp: Removed it

app/src/main/java/com/github/swent/echo/compose/components/TagSelectionDialog.kt
unglazedstamp: Thanks for the suggestion, I will change that

app/src/main/java/com/github/swent/echo/compose/components/TagSelectionDialog.kt
unglazedstamp: I think it will  be confusing to add another Icon, but maybe I don't get it , we can discuss that after the next meeting

ntissieres (approved):
The changes are approved, nice work !

PR #116, opened 2024-04-23
Adapt the event and tag data class to fit the backend

octogradiste:
I added a new `parentId` field to the `Tag`. This is necessary for building the tree structure we want to achieve with the tags.

Also, I simplified the `Event` model. The `creator` is now of type `EventCreator` which has a `userId` and a `name`. The reasoning behind this is, that we don't want to store a full `UserProfile` in the `Event` data class, as the `UserProfile` has many other fields (`semester`, `section`, and `tags`) which are not relevant. Those should be fetched separately by the app if it needs the full profile.


violoncelloCH (approved):
LGTM, thanks - this will simplify querying events from the database and reduce the query complexity!

PR #118, opened 2024-04-23
Fix testRegistrationProcess

octogradiste:


alejandrocalles (approved):
Thanks for working on this fix! It should be all good.

PR #120, opened 2024-04-24
Fix/homescreen modularization

ntissieres:
Modularisation the HomeScreen Component to have the TopAppBar and the hamburger menu in separate files


Greenade (approved):
very nice modularisation ! Will be much more readable and will definitely help me make the viewModel 💯

PR #124, opened 2024-04-24
Feature/home-screen-view-model

Greenade:
created the viewModel fot the homescreen and updated the homescreen and all subclasses with the viewModel


Greenade:
/!\ adding the viewmodel to the homescreen makes the app crash /!\ 
need help please @unglazedstamp  @octogradiste 
I tried basically adding nothing and simply passing the viewmodel as an argument to the home screen in AppNavigationHost but the app crashes instantly.

app/src/main/java/com/github/swent/echo/viewmodels/HomeScreenViewModel.kt
```diff
@@ -0,0 +1,167 @@
...
+            )
+        refreshFiltersContainer()
+    }
+
+    // End of methods to set the filters container values
+
+    private fun filterEvents() {
+        _displayEventList.value =
+            allEventsList.filter { event ->
+                event.tags.any { tag -> filterTagList.any { tag2 -> tag.tagId == tag2.tagId } }
```
octogradiste, line 126: You could use `event.tags.any { tag -> filterTagList.contains(tag) }`. Makes it easier to read... But your solution works as well.

app/src/test/java/com/github/swent/echo/viewmodels/HomeScreenViewModelTest.kt
```diff
@@ -0,0 +1,118 @@
...
+        Dispatchers.setMain(StandardTestDispatcher(scheduler))
+        runBlocking {
+            homeScreenViewModel = HomeScreenViewModel(mockedRepository, fakeAuthenticationService)
+        }
+        scheduler.runCurrent()
+    }
+
+    @Test
+    fun overlayTest() {
+        TestCase.assertEquals(homeScreenViewModel.overlay.value, Overlay.NONE)
```
octogradiste, line 43: I think the standard `assertEquals` should be imported from `import org.junit.Assert.*`... Probably this `assertEquals` from `TestCase` does the same but I think to also be more consistent with the other test we should use the first one.

app/src/main/java/com/github/swent/echo/compose/components/TopAppBar.kt
```diff
@@ -21,15 +22,23 @@ import androidx.compose.ui.res.stringResource
...
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
 
 @OptIn(ExperimentalMaterial3Api::class)
 @Composable
-fun TopAppBar(scope: CoroutineScope, drawerState: DrawerState, mode: MutableState<MapOrListMode>) {
+fun TopAppBar(
+    scope: CoroutineScope,
+    drawerState: DrawerState,
+    homeScreenViewModel: HomeScreenViewModel
```
octogradiste, line 35: I think here the `mode` could be just of type `MapOrListMode` and not `MutableState<MapOrListMode>` makes the composable cleaner (stateless).

octogradiste (changes requested):
Thanks so much for working on this!!! There are some minor things to change but overall I really like your view model :) Great work!!!

The only thing which is not clean/good, is that you're passing down the view model to all the child composables. This is considered bad practice, see the last bullet point from the [docs](https://developer.android.com/topic/libraries/architecture/viewmodel#best-practices). Not doing this, actually makes everything easier: you don't have to change the signature of all other composables, just get the value indside the home screen (using `by ....collectAsState()`) and then provide directly the actual values to the child composables

octogradiste:
One more thing I forgot to add: you have a lot of reformat commits, maybe we could do an interactive rebase and clean up the git history...

app/src/main/java/com/github/swent/echo/viewmodels/HomeScreenViewModel.kt
Greenade: I wrote that previously but it didn't seem to work idk why 🧐

app/src/test/java/com/github/swent/echo/viewmodels/HomeScreenViewModelTest.kt
Greenade: True I'll change that

app/src/main/java/com/github/swent/echo/compose/components/TopAppBar.kt
Greenade: True

app/src/main/java/com/github/swent/echo/viewmodels/HomeScreenViewModel.kt
```diff
@@ -0,0 +1,243 @@
...
+            )
+        )
+    val filtersContainer = _filtersContainer.asStateFlow()
+
+    private val _profileName =
+        MutableStateFlow<String>("John Doe") // placeholder until we have user profiles
+    val profileName = _profileName.asStateFlow()
+    private val _profileClass =
+        MutableStateFlow<String>("IN - BA6") // placeholder until we have user profiles
+    val profileClass = _profileClass.asStateFlow()
```
octogradiste, line 72: `profileName ` and `profileClass` are not yet used. It would be cool if the hamburger menu would use those values.

app/src/main/java/com/github/swent/echo/compose/components/HomeScreen.kt
```diff
@@ -8,118 +8,105 @@ import androidx.compose.material3.ModalNavigationDrawer
...
     val scope = rememberCoroutineScope()
+    val mode by homeScreenViewModel.mode.collectAsState()
+
     // Hamburger menu compose
     ModalNavigationDrawer(
         // Content of the hamburger menu
-        drawerContent = { HamburgerMenuDrawerSheet(navActions, drawerState, scope, {}) },
+        drawerContent = {
+            HamburgerMenuDrawerSheet(navActions, drawerState, scope) {
+                homeScreenViewModel.signOut()
```
octogradiste, line 40: Here, we should call `navActions.navigateTo(Routes.LOGIN)` to redirect the user to the login screen.

app/src/main/java/com/github/swent/echo/compose/components/HomeScreen.kt
```diff
@@ -8,118 +8,105 @@ import androidx.compose.material3.ModalNavigationDrawer
...
+                onDismiss = { homeScreenViewModel.clearOverlay() },
+                searchEntryCallback = { s -> homeScreenViewModel.onSearchEntryChanged(s) },
+                epflCallback = { homeScreenViewModel.onEpflCheckedSwitch() },
+                sectionCallback = { homeScreenViewModel.onSectionCheckedSwitch() },
+                classCallback = { homeScreenViewModel.onClassCheckedSwitch() },
+                pendingCallback = { homeScreenViewModel.onPendingCheckedSwitch() },
+                confirmedCallback = { homeScreenViewModel.onConfirmedCheckedSwitch() },
+                fullCallback = { homeScreenViewModel.onFullCheckedSwitch() },
+                sortByCallback = { sortBy -> homeScreenViewModel.onSortByChanged(sortBy) },
+                resetFiltersCallback = { homeScreenViewModel.resetFiltersContainer() }
```
octogradiste, line 109: Here you could use a simper syntax, e.g. `onDismiss = homeScreenViewModel::clearOverlay`, avoids wrapping functions into functions. This also works for functions with arguments as long as the signature is the same, e.g. `searchEntryCallback = homeScreenViewModel::onSearchEntryChanged`. Note that there are a lot of other places you could use this syntax, but I didn't add a comment every time.

octogradiste (changes requested):
Excellent work! I've not much to add, every thing looks fine. Also, good job on cleaning up the git history :)

app/src/main/java/com/github/swent/echo/viewmodels/HomeScreenViewModel.kt
Greenade: oh yeah I forgot I added those x) I'll change that

app/src/main/java/com/github/swent/echo/compose/components/HomeScreen.kt
Greenade: makes sense

app/src/main/java/com/github/swent/echo/compose/components/HomeScreen.kt
Greenade: oh that's good I'll use that

octogradiste (approved):

PR #125, opened 2024-04-24
Feature/list mode

ntissieres:
Implementation of the list mode, to display the events on a scrollable list instead of a map

### DIFF
--- a/app/src/main/java/com/github/swent/echo/compose/components/ListDrawer.kt
+++ b/app/src/main/java/com/github/swent/echo/compose/components/ListDrawer.kt
@@ -18,38 +18,46 @@ import androidx.compose.material3.Button
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.getValue
+import androidx.compose.runtime.MutableState
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
-import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.unit.dp
+import com.github.swent.echo.R
 import com.github.swent.echo.data.model.Event
 import java.time.format.DateTimeFormatter
 
 @Composable
 fun ListDrawer(eventsList: List<Event>) {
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/github/swent/echo/compose/components/ListDrawer.kt
```diff
@@ -0,0 +1,148 @@
...
+        Row(
+            modifier = Modifier.height(60.dp).testTag("list_event_row_${event.eventId}"),
+            horizontalArrangement = Arrangement.Center,
+            verticalAlignment = Alignment.CenterVertically,
+        ) {
+            val paddingItems = 3.dp
+            val widthSmallItems = 50.dp
+            val widthLargeItems = 95.dp
+            Text(
+                // We should add a way to get the childest tag from event.tags . For now, I use
```
srsingh04, line 73: knitpick: in the comments, you meant children tag instead childest tag??

app/src/test/java/com/github/swent/echo/compose/components/ListDrawerTest.kt
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    @Test
+    fun testAllEventsDisplayedCorrectly() {
+        // Check max 5 events to avoid interminable test when we will have a lot of events
+        for (i in 0..min(SAMPLE_EVENTS.size - 1, 4)) {
+            composeTestRule
+                .onNodeWithTag("list_event_item_${SAMPLE_EVENTS[i].eventId}")
+                .assertExists()
+            composeTestRule
```
srsingh04, line 37: Knitpick: It could be good practice to wrap the duplicate code in the loop into a function to avoid code duplication and a cleaner code.

srsingh04 (approved):

app/src/main/java/com/github/swent/echo/compose/components/ListDrawer.kt
ntissieres: Yes, childest is not a true word haha, I meant the last child of the tags hierarchy. But this comment will be removed after this problem is solved

app/src/test/java/com/github/swent/echo/compose/components/ListDrawerTest.kt
ntissieres: I'm not sure to see which duplication you are talking about ?

srsingh04: I meant wrapping the code inside the for loop in a separate function and then call the following function while testing for 5 sample events. In this way, if you would ever want to modify the tests, it would be easier to do when the logic is wrapped in a function.

PR #137, opened 2024-04-25
Move all compose tests (which work with robolectric) to the test folder

### LINKED ISSUE #108: Use robolectric to test our composables
Go see the [wiki](https://github.com/echo-swent/echo/wiki/Tests) on how to do this.

octogradiste:
This should make our pipeline faster. :)
Closes #108.


octogradiste:
Actually it doesn't make the CI faster, as it still has to start an emulator for the remaining android tests. But using robolectric makes running the tests locally much faster, so it still makes sense to use it.

unglazedstamp (approved):
Looks good

PR #140, opened 2024-04-26
Feature/profile creation 

srsingh04:
Profile_creation_composable which lets the user enter their personal information after logging in or also to modify it after the account has been created before.

### DIFF
--- a/app/src/androidTest/java/com/github/swent/echo/compose/authentication/createProfileTest.kt
+++ b/app/src/androidTest/java/com/github/swent/echo/compose/authentication/createProfileTest.kt
@@ -1,0 +1,72 @@
+package com.github.swent.echo.compose.authentication
+
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.test.junit4.createComposeRule
+import androidx.compose.ui.test.onAllNodesWithContentDescription
+import androidx.compose.ui.test.onNodeWithTag
+import androidx.compose.ui.test.performClick
+import androidx.navigation.compose.rememberNavController
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import com.github.swent.echo.data.model.SectionEPFL
+import com.github.swent.echo.data.model.SemesterEPFL
+import com.github.swent.echo.data.model.Tag
+import com.github.swent.echo.ui.navigation.NavigationActions
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+class CreateProfileTest {
+    @get:Rule val composeTestRule = createComposeRule()
+
+    @Test
+    fun profileCreationScreen_Loads() {
+        // Verify that the Profile Creation screen loads successfully
+        composeTestRule.setContent {
+            ProfileCreationUI(
+                listOf(SectionEPFL.AR, SectionEPFL.IN, SectionEPFL.SC, SectionEPFL.GM),
+                listOf(SemesterEPFL.BA1, SemesterEPFL.BA2),
+                listOf(Tag("1", "Sports"), Tag("2", "Music")),
+                navAction = NavigationActions(navController = rememberNavController())
+            )
+        }
+        // Assert that certain elements are present on the screen
+        composeTestRule.onNodeWithTag("Save").assertExists()
+        composeTestRule.onNodeWithTag("Back").assertExists()
+        composeTestRule.onNodeWithTag("AddTag").assertExists()
+        composeTestRule.onNodeWithTag("FirstName").assertExists()
+        composeTestRule.onNodeWithTag("LastName").assertExists()
+        composeTestRule.onNodeWithTag("Music").assertExists()
+        composeTestRule.onNodeWithTag("Sports").assertExists()
+        composeTestRule.onNodeWithTag("Section").assertExists()
+        composeTestRule.onNodeWithTag("Semester").assertExists()
+        composeTestRule.onAllNodesWithContentDescription("list dropdown")[0].performClick()
+        // Verify that the dropdown menu appears
+        composeTestRule.onNodeWithTag("SC").assertExists()
+        composeTestRule.onNodeWithTag("IN").assertExists()
+        // composeTestRule.onNodeWithTag("Math").assertExists()
+        composeTestRule.onNodeWithTag("GM").assertExists()
+        composeTestRule.onAllNodesWithContentDescription("list dropdown")[1].performClick()
+        composeTestRule.onNodeWithTag("BA1").assertExists()
+        composeTestRule.onNodeWithTag("BA2").assertExists()
+    }
+    /*
+       @Test
+       fun profileCreationScreen_SaveButtonClicked() {
+           var saveButtonClicked = false
+           val saveOnClick: () -> Unit = { saveButtonClicked = true }
+           composeTestRule.setContent {
+               ProfileCreationUI(
+                   emptyList(),
+                   emptyList(),
+                   emptyList(),
+                   navAction = NavigationActions(navController = rememberNavController())
+               )
+           }
+           // Simulate a click on the Save button
+           composeTestRule.onNodeWithText("Save").performClick()
+           assert(saveButtonClicked)
+       }
+    */
+}
--- a/app/src/main/java/com/github/swent/echo/compose/authentication/createProfile.kt
+++ b/app/src/main/java/com/github/swent/echo/compose/authentication/createProfile.kt
@@ -1,0 +1,251 @@
+package com.github.swent.echo.compose.authentication
+
+import androidx.compose.foundation.clickable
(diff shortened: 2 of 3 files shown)


app/src/main/java/com/github/swent/echo/compose/authentication/createProfile.kt
```diff
@@ -0,0 +1,251 @@
...
+/**
+ * A composable function that displays the UI for creating a user profile.
+ *
+ * @param sectionList The list of sections to be displayed in the dropdown menu.
+ * @param semList The list of semesters to be displayed in the dropdown menu.
+ * @param tagList The list of tags to be displayed as chips.
+ */
+@Composable
+fun ProfileCreationUI(
+    sectionList: List<Section>,
```
ntissieres, line 72: I think those List are temporary, they will be fetched from the repository later. Instead of asking them as arguments for the ProfileCreationUI, you could add samples outside the function and call them when needed

app/src/main/java/com/github/swent/echo/compose/authentication/createProfile.kt
```diff
@@ -0,0 +1,251 @@
...
+    // viewModel: createProfileViewModel = hiltViewModel(),
+    navAction: NavigationActions
+) {
+    Box(modifier = Modifier.fillMaxSize().padding(16.dp), contentAlignment = Alignment.Center) {
+        Column(modifier = Modifier.fillMaxSize().padding(16.dp)) {
+            var firstName by remember { mutableStateOf("") }
+            var lastName by remember { mutableStateOf("") }
+
+            // Back button
+            Icon(
```
ntissieres, line 84: It would make more sense to use an IconButton : https://www.composables.com/material/iconbutton

app/src/main/java/com/github/swent/echo/compose/authentication/createProfile.kt
```diff
@@ -0,0 +1,251 @@
...
+            Text(
+                stringResource(id = R.string.profile_creation_tags),
+                fontSize = 20.sp,
+                fontWeight = FontWeight.Bold
+            )
+
+            Spacer(modifier = Modifier.height(10.dp))
+
+            Row() {
+                for (tag in tagList) {
```
ntissieres, line 127: You can use Yoan implementation of tag selection. You need to rebase your branch and then use TagSelectionDialog. Ask Yoan if you need help on how to use it

app/src/main/java/com/github/swent/echo/compose/authentication/createProfile.kt
```diff
@@ -0,0 +1,251 @@
...
+                    InputChipFun(tag.name) {}
+                    Spacer(modifier = Modifier.width(5.dp))
+                }
+                Spacer(modifier = Modifier.width(5.dp))
+
+                // Add tag button
+                SmallFloatingActionButton(
+                    onClick = {},
```
ntissieres, line 135: You can use Yoan implementation of tag selection into the onCLick. You need to rebase your branch and then use TagSelectionDialog. Ask Yoan if you need help on how to use it

app/src/main/java/com/github/swent/echo/compose/authentication/createProfile.kt
```diff
@@ -0,0 +1,251 @@
...
+                    containerColor = MaterialTheme.colorScheme.secondaryContainer,
```
ntissieres, line 136: Try to avoid forcing color except if it's realy necessary. Normally, everything is done automatically for the color

app/src/main/java/com/github/swent/echo/compose/authentication/createProfile.kt
```diff
@@ -0,0 +1,251 @@
...
+                modifier = Modifier.fillMaxWidth().testTag("Save")
+            ) {
+                Text(text = stringResource(id = R.string.profile_creation_save_button))
+            }
+        }
+    }
+}
+
+@Composable
+fun DropDownListFunctionWrapper(elementList: List<Any>, label: Int) {
```
ntissieres, line 158: Good DropDown menu ;)

app/src/main/java/com/github/swent/echo/compose/authentication/createProfile.kt
```diff
@@ -0,0 +1,251 @@
...
+                            showDropdown = false
+                        },
+                        modifier = Modifier.testTag(elem.toString())
+                    )
+                }
+            }
+        }
+    }
+}
+
```
ntissieres, line 204: Don't forget to comment what each Composable are doing, this one wasn't really clear at a first sight

app/src/androidTest/java/com/github/swent/echo/compose/authentication/createProfileTest.kt
```diff
@@ -0,0 +1,72 @@
...
+import com.github.swent.echo.ui.navigation.NavigationActions
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+class CreateProfileTest {
+    @get:Rule val composeTestRule = createComposeRule()
+
+    @Test
```
ntissieres, line 23: Before adding your @Test, you should define your composeTestRule.setContent inside a @Before fun setup() function in order to instantiate your composable before each @Test. Take a look at the test done by the other to have examples of the correct structure.

app/src/androidTest/java/com/github/swent/echo/compose/authentication/createProfileTest.kt
```diff
@@ -0,0 +1,72 @@
...
+        // Verify that the Profile Creation screen loads successfully
+        composeTestRule.setContent {
+            ProfileCreationUI(
+                listOf(SectionEPFL.AR, SectionEPFL.IN, SectionEPFL.SC, SectionEPFL.GM),
+                listOf(SemesterEPFL.BA1, SemesterEPFL.BA2),
+                listOf(Tag("1", "Sports"), Tag("2", "Music")),
+                navAction = NavigationActions(navController = rememberNavController())
+            )
+        }
+        // Assert that certain elements are present on the screen
```
ntissieres, line 34: It's better to separate each test instead of having a big Test function that test everything

app/src/main/java/com/github/swent/echo/compose/authentication/createProfile.kt
```diff
@@ -0,0 +1,251 @@
...
+                offset = DpOffset(0.dp, 0.dp)
+            ) {
+                elementList.forEach { elem ->
+                    DropdownMenuItem(
+                        text = { Text(elem.toString()) },
+                        onClick = {
+                            selectedField = elem.toString()
+                            showDropdown = false
+                        },
+                        modifier = Modifier.testTag(elem.toString())
```
ntissieres, line 197: use testTag("create_profile_${elem.toString()}") instead

app/src/androidTest/java/com/github/swent/echo/compose/authentication/createProfileTest.kt
```diff
@@ -0,0 +1,72 @@
...
+        composeTestRule.onNodeWithTag("AddTag").assertExists()
+        composeTestRule.onNodeWithTag("FirstName").assertExists()
+        composeTestRule.onNodeWithTag("LastName").assertExists()
+        composeTestRule.onNodeWithTag("Music").assertExists()
+        composeTestRule.onNodeWithTag("Sports").assertExists()
+        composeTestRule.onNodeWithTag("Section").assertExists()
+        composeTestRule.onNodeWithTag("Semester").assertExists()
+        composeTestRule.onAllNodesWithContentDescription("list dropdown")[0].performClick()
+        // Verify that the dropdown menu appears
+        composeTestRule.onNodeWithTag("SC").assertExists()
```
ntissieres, line 46: Try to make a loop to test those tags, it avoid redondance and the test needs to work with a general list of section and not those specific values. You can check how i did it on ListDrawerTest if you need an example on how to do it

ntissieres (changes requested):
The code still needs some improvements :
- Files name starts with a capital letter.  Change the name "createProfile" and "createProfileTest" to "CreateProfile" and "CreateProfileTest"
- Try to give more unique name for the testTags, because for example multiple files could contain a composable with tag "back". Name it instead "create_profile_back"
- Correct the individual comment i posted

app/src/main/java/com/github/swent/echo/compose/authentication/createProfile.kt
octogradiste: We actually already have those values in the code. Just use `SectionEPFL.values()` (same for `SemesterEPFL`).

app/src/androidTest/java/com/github/swent/echo/compose/authentication/createProfileTest.kt
srsingh04: I didn't put it in a @before because there was only 1 test case XD

alejandrocalles:
Duplicate of #188.

PR #141, opened 2024-04-26
integrate the tag selection dialog into the event creation screen

unglazedstamp:
based on #113

### DIFF
--- a/app/src/androidTest/java/com/github/swent/echo/compose/event/EventScreenTest.kt
+++ b/app/src/androidTest/java/com/github/swent/echo/compose/event/EventScreenTest.kt
@@ -1,36 +1,44 @@
 package com.github.swent.echo.compose.event
 
+import androidx.activity.compose.setContent
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.test.assertIsDisplayed
 import androidx.compose.ui.test.assertTextContains
 import androidx.compose.ui.test.assertTextEquals
-import androidx.compose.ui.test.junit4.createComposeRule
+import androidx.compose.ui.test.junit4.createAndroidComposeRule
 import androidx.compose.ui.test.onNodeWithTag
 import androidx.compose.ui.test.performClick
 import androidx.compose.ui.test.performScrollTo
 import androidx.compose.ui.test.performTextInput
 import androidx.compose.ui.test.performTextReplacement
 import androidx.lifecycle.SavedStateHandle
+import com.github.swent.echo.MainActivity
 import com.github.swent.echo.R
 import com.github.swent.echo.authentication.AuthenticationService
 import com.github.swent.echo.data.model.Event
 import com.github.swent.echo.data.model.Tag
 import com.github.swent.echo.data.repository.Repository
 import com.github.swent.echo.viewmodels.event.EventViewModel
+import dagger.hilt.android.testing.HiltAndroidRule
+import dagger.hilt.android.testing.HiltAndroidTest
 import io.mockk.mockk
+import junit.framework.TestCase.assertTrue
 import org.junit.Assert
 import org.junit.Before
 import org.junit.Ignore
 import org.junit.Rule
 import org.junit.Test
 
+@HiltAndroidTest
 class EventScreenTest {
 
     private val STRING_1 = "a test string"
     private val STRING_2 = "another test string"
     private val STRING_3 = "a third test string"
 
-    @get:Rule val composeTestRule = createComposeRule()
+    @get:Rule(order = 0) val hiltRule = HiltAndroidRule(this)
+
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/github/swent/echo/compose/event/EventTagEntry.kt
```diff
@@ -1,42 +1,72 @@
...
+                    )
+                }
             }
+            OutlinedIconButton(
+                modifier = Modifier.testTag("add-tag-button"),
+                onClick = { dialogVisible = true }
+            ) {
+                val icon = Icons.Filled.Add
+                Icon(imageVector = icon, contentDescription = icon.name)
+            }
```
violoncelloCH, line 59: the button to add a new tag should also have a label next to the plus icon; at least "Add" or better "Add Tag" to make the button accessible for screen reader users:
image

app/src/main/java/com/github/swent/echo/compose/event/EventTagEntry.kt
```diff
@@ -1,42 +1,72 @@
...
-                Text(tag.name)
-                Icon(
-                    imageVector = Icons.Filled.Clear,
-                    contentDescription =
-                        stringResource(R.string.edit_event_screen_delete_tag_button) + tag.name
-                )
+    var dialogVisible by remember { mutableStateOf(false) }
+    Column(modifier = Modifier.fillMaxWidth().padding(EVENT_PADDING_BETWEEN_INPUTS)) {
+        EventEntryName(stringResource(R.string.edit_event_screen_tags))
+        Row(modifier = Modifier.horizontalScroll(rememberScrollState())) {
```
violoncelloCH, line 39: I don't think using horizontal scroll on this makes sense from a UX perspective. It's very easy to accidentally click one of the tags when scrolling and therefore delete it. Couldn't that simply expand into multiple rows?

app/src/main/java/com/github/swent/echo/compose/event/EventTagEntry.kt
```diff
@@ -1,42 +1,72 @@
...
+                    onClick = { onTagDeleted(tag) },
+                    modifier = Modifier.testTag("${tag.name}-tag-button")
+                ) {
+                    Text(tag.name)
+                    Icon(
+                        imageVector = Icons.Filled.Clear,
+                        contentDescription =
+                            stringResource(R.string.edit_event_screen_delete_tag_button) + tag.name
+                    )
+                }
```
violoncelloCH, line 51: Did you look into whether you could use [input chips](https://developer.android.com/develop/ui/compose/components/chip#input) for this ?
At the very least, the margin between the Text and the X icon should be adjusted to match the chip component.

violoncelloCH (changes requested):
Thanks for the incorporation of the tag select screen into the event creation screen! I've just got some minor, mostly design related comments :)

app/src/androidTest/java/com/github/swent/echo/compose/event/EventScreenTest.kt
```diff
@@ -1,47 +1,60 @@
...
+
+    @get:Rule(order = 1) val composeTestRule = createAndroidComposeRule<MainActivity>()
 
     private val mockedRepository = mockk<Repository>(relaxed = true)
     private val mockedAuthenticationService = mockk<AuthenticationService>(relaxed = true)
     private lateinit var eventViewModel: EventViewModel
     private val savedEventId = SavedStateHandle(mapOf())
 
+    // @Inject var tagViewModel: TagViewModel = TagViewModel(mockedRepository)
+
```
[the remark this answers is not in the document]
unglazedstamp: Yes, thanks

app/src/main/java/com/github/swent/echo/compose/event/EventTagEntry.kt
unglazedstamp: That's the role of the `contentDescription`, but it's true that the icon description isn't very descriptive

app/src/main/java/com/github/swent/echo/compose/event/EventTagEntry.kt
unglazedstamp: I was afraid that too long or too many tags could fill the UI, but after trying it, I see that it makes more sense to have multiple rows.

app/src/main/java/com/github/swent/echo/compose/event/EventTagEntry.kt
unglazedstamp: Thank you for finding input chips, it's exactly the use case I need for the tags.

violoncelloCH (approved):

PR #154, opened 2024-04-26
Delete existing markers when the list of events is updated.

### LINKED ISSUE #149: Map bugfix events are not refreshed
When the list of events passed as a map updates, new events may appear, but are not clickable. And when events should disapear the marker stays but is not clickable.

alejandrocalles:
Closes #149.


Greenade (approved):
very nice ! this fixes the bug that was visible while filtering events in the homescreen viewmodel 👍

PR #157, opened 2024-04-26
add userProfile <-> association relationship (committee member & association subscription)

### LINKED ISSUE #104: allow for userProfiles to be associated to associations (i.e. members of comittee & association subscriptions)
- [x] create table for the relation
- [x] fetch the relation as part of the userProfile data class
- [x] ignore on setter for committee member attribute (user doesn't have the right to add relations himself)

violoncelloCH:
closes #104 

needs #109 to be merged first


app/src/main/java/com/github/swent/echo/data/model/UserProfile.kt
```diff
@@ -5,7 +5,9 @@ data class UserProfile(
     val name: String,
     val semester: Semester?,
     val section: Section?,
-    val tags: Set<Tag>
+    val tags: Set<Tag>,
+    val committeeMember: Set<Association>,
+    val associationsSubscriptions: Set<Association>,
```
octogradiste, line 10: You could add a default value, e.g. `comitteeMember: Set<Association> = emptySet()`. This way you don't need to change code all over the place. 

One other possibility is to not add this to the `UserProfile`, but as functionality to the `Repository` with functions like `getComitteMember(userId: String)`.

octogradiste (commented):
Some thoughts on the `UserProfile`.

app/src/main/java/com/github/swent/echo/data/supabase/SupabaseDataSource.kt
```diff
@@ -86,5 +87,11 @@ class SupabaseDataSource(private val supabase: SupabaseClient) : RemoteDataSourc
         val userTagsSupabase =
             userProfile.tags.map { tag -> UserTagSupabase(userProfile.userId, tag.tagId) }
         supabase.from("user_tags").upsert(userTagsSupabase)
+
+        val associationSubscriptionSupabase =
+            userProfile.associationsSubscriptions.map { association ->
+                AssociationMemberSupabase(userProfile.userId, association.associationId)
```
unglazedstamp, line 93: It could be helpful to explain here that `AssociationMemberSupabase` represents an association subscription and not a membership either with a comment or with an additional class named `AssociationSubscriptionSupabase`

unglazedstamp (approved):
Very clean code, some additional comments could be helpful to understand it faster.

app/src/main/java/com/github/swent/echo/data/model/UserProfile.kt
octogradiste: After discussing this in person with @violoncelloCH and @unglazedstamp, it makes more sense to add those to the `UserProfile` instead of adding getters to the `Repository`. Also, as pointed out by @violoncelloCH a default value doesn't really make sense here. So, it's better to adapt the code everywhere...

violoncelloCH: exactly ! thanks for writing up the conclusion here already :)

app/src/main/java/com/github/swent/echo/data/supabase/SupabaseDataSource.kt
violoncelloCH: you're right, I missed to rename that class; it should indeed be named `AssociationSubscriptionSupabase`...
I've changed that

unglazedstamp (approved):
Thank you for the improvement !

PR #159, opened 2024-04-27
App Logo

Greenade:
Created and added the logo as an icon and as a rounded icon


violoncelloCH (commented):
As mentioned on the chat, i think that the megaphone icon looks a bit disbalanced whit the Text at the current position. (The above part loogs higher than the below one).

I'd try two variants:
- add at least some margin between the lowest tip and the tip of the megaphone and the top line of the text
- make the text completely clash with the icon, i.e. make the two touch voluntarily (obviously different versions are possible here; needs some playing around and it's also possible that it does not work at all)

violoncelloCH (approved):
thanks for the updated version :)

PR #160, opened 2024-04-27
Add name attribute to semester

octogradiste:


Greenade (approved):
nice, will be usefull for the home screen view model later 👍

PR #164, opened 2024-04-29
Feature: add participant limit to the event screen

unglazedstamp:


app/src/main/java/com/github/swent/echo/compose/event/EventScreen.kt
```diff
@@ -27,7 +30,13 @@ fun EventScreen(
     onEventBackButtonPressed: () -> Unit,
     eventViewModel: EventViewModel
 ) {
-    Column(modifier = Modifier.verticalScroll(rememberScrollState())) {
+    val focusManager = LocalFocusManager.current
+    Column(
+        modifier =
+            Modifier.verticalScroll(rememberScrollState()).pointerInput(Unit) {
```
ntissieres, line 36: Why this over LazyColumn ?

ntissieres (approved):
Good code !

app/src/main/java/com/github/swent/echo/compose/event/EventScreen.kt
unglazedstamp: LazyColumn should be used when there is an unknown (potentially large) number of items. In this code there is a known number of items (< 10), so It's not necessary to use LazyColumn

PR #167, opened 2024-04-29
Extend the local data source to support better syncing strategies

octogradiste:


violoncelloCH (commented):
This looks good to me. It will allow me to implement a more sophisticated synchronisation strategy for #193 ! So please continue on this one :)

violoncelloCH (approved):
LGTM. 
I'm just wondering rn if it might make sense to also have a way to know which elements are being deleted (since out of date), so they could be refetched from Supabase if needed. The delete methods could return a list of id's of the deleted events.

octogradiste:
Yeah, makes sense, I'll check how to do this with room and update the interface for the actual implementation!

PR #169, opened 2024-04-29
Feature/graphic update 2 event info sheet

Greenade:
changed the graphical layout of the event info sheet, null organizer now doesn't crash and displays creator name instead.

### DIFF
--- a/app/src/main/java/com/github/swent/echo/compose/components/EventInfoSheet.kt
+++ b/app/src/main/java/com/github/swent/echo/compose/components/EventInfoSheet.kt
@@ -62,9 +62,7 @@ fun EventInfoSheet(
     val context = LocalContext.current
 
     ModalBottomSheet(
-        modifier = Modifier
-            .fillMaxSize()
-            .testTag("event_info_sheet"),
+        modifier = Modifier.fillMaxSize().testTag("event_info_sheet"),
         onDismissRequest = onDismiss,
         sheetState = sheetState,
     ) {
@@ -72,10 +70,9 @@ fun EventInfoSheet(
 
         Box(
             modifier =
-            Modifier
-                .fillMaxWidth()
-                .fillMaxHeight(0.5f)
-                .padding(start = 20.dp, end = 20.dp, top = 0.dp, bottom = 32.dp)
+                Modifier.fillMaxWidth()
+                    .fillMaxHeight(0.5f)
+                    .padding(start = 20.dp, end = 20.dp, top = 0.dp, bottom = 32.dp)
         ) {
             Text(
                 text = event.title,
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/github/swent/echo/compose/components/EventInfoSheet.kt
```diff
@@ -64,14 +63,21 @@ fun EventInfoSheet(
     val hour = event.startDate.hour
     val minute = event.startDate.minute
     val displayMonth = if (month < 10) "0$month" else month.toString()
-    val displayDate = "$day/$displayMonth\n$hour:$minute"
+    val displayMinute = if (minute < 10) "0$minute" else minute.toString()
+    val displayDay = if (day < 10) "0$day" else day.toString()
+    val displayHour = if (hour < 10) "0$hour" else hour.toString()
+    val displayDate = "$displayDay/$displayMonth\n$displayHour:$displayMinute"
```
octogradiste, line 69: Instead of doing this by hand you could use a `DateTimeFormatter` as follows: `val formatter = DateTimeFormatter.ofPattern("dd/MM\nHH:mm")` and then simply `val displayDate = formatter.format(event.startDate)`

app/src/main/java/com/github/swent/echo/compose/components/EventInfoSheet.kt
```diff
@@ -88,11 +94,12 @@ fun EventInfoSheet(
             )
             Text(
                 modifier = Modifier.padding(top = 24.dp),
-                text = event.organizer!!.name,
+                text = event.organizer?.name ?: event.creator.name,
                 style =
                     TextStyle(
                         fontSize = 20.sp,
                         fontWeight = FontWeight(600),
+                        color = MaterialTheme.colorScheme.secondary
```
octogradiste, line 102: Instead of defining a custom text style, we could use the one defined by `MaterialTheme`, e.g. `style = MaterialTheme.typography.titleMedium.copy(color = MaterialTheme.colorScheme.secondary)`.

app/src/main/java/com/github/swent/echo/compose/components/EventInfoSheet.kt
```diff
@@ -102,6 +109,7 @@ fun EventInfoSheet(
...
```
octogradiste, line 112: Same thing here, e.g. `style = MaterialTheme.typography.labelMedium.copy(color = MaterialTheme.colorScheme.secondary)`.

app/src/main/java/com/github/swent/echo/compose/components/EventInfoSheet.kt
```diff
@@ -114,19 +122,54 @@ fun EventInfoSheet(
                     )
             )
```
octogradiste, line 123: Here we could use for example `style = MaterialTheme.typography.bodyLarge`.

octogradiste (changes requested):
Good update, maybe we could use the text styles defined `MaterialTheme` instead of hardcoding them. What do you thinkg?

octogradiste:
Also, I just checked the analysis from Sonar Cloud, and it complains about a callback which is never used `onShowPeopleButtonPressed`. If it's actually the case I think we should just remove it.

app/src/main/java/com/github/swent/echo/compose/components/EventInfoSheet.kt
Greenade: oh that's cool I'll use that

octogradiste (approved):
Thanks for the changes! LGTM

PR #170, opened 2024-04-29
Always use the `SimpleRepository` in android tests.

octogradiste:
This makes sure that once we change to the real `Repository` implementation, our android tests won't break/be flaky.


unglazedstamp (approved):
LGTM

Issue #163, opened 2024-04-29
App crashes when clicking on the compass icon at the top right of the map screen


violoncelloCH:
the following output log output in logcat:
```
2024-04-29 10:39:56.055 31490-31530 EGL_emulation           com.github.swent.echo                D  app_time_stats: avg=1940.89ms min=31.69ms max=8906.58ms count=5
2024-04-29 10:39:56.520 31490-31530 libc                    com.github.swent.echo                A  Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x30 in tid 31530 (GLThread 96), pid 31490 (thub.swent.echo)
2024-04-29 10:39:56.633 31490-31510 EGL_emulation           com.github.swent.echo                D  app_time_stats: avg=1973.08ms min=3.07ms max=11188.98ms count=6
2024-04-29 10:39:56.760 31571-31571 DEBUG                   pid-31571                            A  Cmdline: com.github.swent.echo
2024-04-29 10:39:56.760 31571-31571 DEBUG                   pid-31571                            A  pid: 31490, tid: 31530, name: GLThread 96  >>> com.github.swent.echo <<<
2024-04-29 10:39:56.761 31571-31571 DEBUG                   pid-31571                            A        #01 pc 000000000089ee90  /data/app/~~184vFka7x-GS94onzKuRzg==/com.github.swent.echo-R1eThci6L2zgOJz_hoiShw==/lib/x86_64/libmapbox-gl.so (BuildId: 4390bec0d101ed6a55157c48a4fa4f183581be86)
2024-04-29 10:39:56.762 31571-31571 DEBUG                   pid-31571                            A        #02 pc 00000000008a2308  /data/app/~~184vFka7x-GS94onzKuRzg==/com.github.swent.echo-R1eThci6L2zgOJz_hoiShw==/lib/x86_64/libmapbox-gl.so (BuildId: 4390bec0d101ed6a55157c48a4fa4f183581be86)
2024-04-29 10:39:56.762 31571-31571 DEBUG                   pid-31571                            A        #03 pc 00000000008a486b  /data/app/~~184vFka7x-GS94onzKuRzg==/com.github.swent.echo-R1eThci6L2zgOJz_hoiShw==/lib/x86_64/libmapbox-gl.so (BuildId: 4390bec0d101ed6a55157c48a4fa4f183581be86)
[...]

PR #172, opened 2024-04-30
Display the location and fix tests.

### LINKED ISSUE #89: Display the location of the user on the map.
Display the location of the user on the map.
Should request permission from the user to access the location.
Should be able to center the map on the location.

### LINKED ISSUE #127: Fix `MapDrawerAndroidTest`.
Should instantiate the provider directly instead of relying on injeciton from hilt.
Clean up unnecessary tests, and remove exposed `mapView` fields from the `OsmdroidMapViewProvider`.

### LINKED ISSUE #129: Default view of the map on the EPFL
Set the map default zoom and position centered on the EPFL campus

### LINKED ISSUE #178: Fix Map not showing after navigating to event list and back.

alejandrocalles:
Closes #89.
Closes #127.
Closes #129.
Closes #178.

### DIFF
--- a/app/src/main/java/com/github/swent/echo/compose/map/Map.kt
+++ b/app/src/main/java/com/github/swent/echo/compose/map/Map.kt
@@ -39,11 +39,7 @@ fun <T : View> EchoAndroidView(
     AndroidView(
         modifier = modifier.testTag("mapAndroidView"),
         factory = { factory(it, withLocation) { trigger = true } },
-        update = {
-            if (trigger) {
-                update(it, events, callback, withLocation)
-            }
-        }
+        update = { update(it, events, callback, trigger && withLocation) }
     )
 }
 
@@ -72,17 +68,19 @@ fun MapDrawer(
     mapDrawerViewModel: MapDrawerViewModel = hiltViewModel(),
 ) {
     var displayLocation by remember { mutableStateOf(false) }
+    var permissionWasDenied = false
     val launcher =
         rememberLauncherForActivityResult(ActivityResultContracts.RequestMultiplePermissions()) { p
             ->
-            displayLocation = displayLocation || p.values.any { it }
+            permissionWasDenied = !p.values.any { it }
+            displayLocation = displayLocation || !permissionWasDenied
         }
     val c = LocalContext.current
     /*
     This has to be blocking as we don't want the `EchoAndroidView` to be
     created before launching the permission request.
     */
-    if (runBlocking { permissionsDenied(c) }) {
+    if (!permissionWasDenied && runBlocking { permissionsDenied(c) }) {
         SideEffect { launcher.launch(PERMISSIONS) }
     } else {
         SideEffect { displayLocation = true }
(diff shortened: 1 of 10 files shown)


alejandrocalles:
Closes #127.

alejandrocalles:
This PR solves a bunch of different but related issues.

The main one is #89, but solving #127 was needed to get the test coverage and #129 was too short to get a PR of its own.

alejandrocalles:
I tried to fix #163 to no avail. For the moment, i have set the initial bearing to 0.0 and completely disabled rotation.

Greenade (changes requested):
The idea is good and the code looks great, the position blue thing looks good as well! There are some problems though : 
When we give the permission to display the position, it doesn't display it until we swap pages and come back, so I think you should refresh the map when receiving the permission to display the position. 
Another thing is that for #163 you made it worse x) at least on the emulator when we barely spin the map it crashes now, It doesn't do that on my phone though (doesn't even crash when clicking the compass). 
One last thing is I think the map should be centered arount the position when given the permission.

Greenade (dismissed):

Greenade:
I removed my requested changes as I cannot review your branch :/ please ask somebody to verify it works though 🥲

app/src/main/java/com/github/swent/echo/compose/map/Map.kt
```diff
@@ -22,11 +34,45 @@ fun <T : View> EchoAndroidView(
...
 ) {
-    AndroidView(
-        modifier = modifier.testTag("mapAndroidView"),
-        factory = factory,
-        update = { update(it, events, callback) }
+    val permissions = arrayOf(
+        Manifest.permission.ACCESS_COARSE_LOCATION,
+        Manifest.permission.ACCESS_FINE_LOCATION
     )
+    fun permissionsDenied(context: Context) = permissions.any {
```
unglazedstamp: why is this function declared inside the EchoAndroidView function ?

unglazedstamp (changes requested):
Good code ! I found problems by running the app:
1. If I refuse location permission it ask it again when I click on an event pin, it only does it the first time but it shouldn't at all
2. Some pins in the background becomes black when I click on one of them
3. If I allow approximate location, go to the list and come back to the map, it asks again if I want precise location
The second issue is not necessarly related to this PR and none of them are not blocking the app, so if you have the time to solve 1. and 3. it would be nice, otherwise we can merge this PR and solve them later.

app/src/main/java/com/github/swent/echo/compose/map/Map.kt
alejandrocalles: Good catch! I've moved it out.

unglazedstamp (approved):

unglazedstamp:
All the problem are solved when I run the app on a real phone instead of the emulator. Thanks a lot for all this work !

unglazedstamp (approved):

PR #173, opened 2024-04-30
Feature/time filter

ntissieres:
Adding a time slider to filter the events according to their start and end date

### DIFF
--- a/app/src/main/java/com/github/swent/echo/compose/components/searchmenu/SearchMenuFilters.kt
+++ b/app/src/main/java/com/github/swent/echo/compose/components/searchmenu/SearchMenuFilters.kt
@@ -8,6 +8,7 @@ import androidx.compose.foundation.layout.absoluteOffset
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.offset
 import androidx.compose.foundation.layout.width
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material.icons.Icons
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/github/swent/echo/compose/components/searchmenu/SearchMenuFilters.kt
```diff
@@ -205,9 +210,73 @@ fun SortByDisplayer(sortBy: SortBy, sortByCallback: (SortBy) -> Unit) {
...
+    Column(modifier = Modifier.width(350.dp)) {
+        // Slider for the date range
+        RangeSlider(
+            value = sliderPosition,
+            steps = 13,
+            onValueChange = { range ->
+                sliderPosition = range
+                timeSliderCallback(sliderPosition.start, sliderPosition.endInclusive)
+            },
+            valueRange = 0f..14f,
```
Greenade, line 226: only 14 days range for the search ? that seems a bit low to me

Greenade (changes requested):
Looks amazing ! Thank you for the hard work, the use of the viewModel is very good as well. The only thing I have to say is that I think 2 weeks range is too low, we should at least be able to view event a month in advance in my opinion. Other than that it's perfect !

app/src/main/java/com/github/swent/echo/compose/components/searchmenu/SearchMenuFilters.kt
ntissieres: That's a good point, especially for associations that can plan events months in advance. But at the same time, I think users that will use this feature will be way more interested in events on the next two weeks than in plannifying their whole calendar. But, the problem is actually that we can't display events if they are planned more than 2 weeks ahead. I can change that and display all events that will take place 2 weeks+ if the user have the right side of the slider full on the right.
It's more about place management, the slider fits perfectly in the search sheet where 2 boxes to select the dates are too big and not as practical. Having a 30 steps slider will also be a pain to select a specific day and be accurate with the finger.

Greenade: that's a good solution ! you just need to display that clearly so thet it's not confusing :)

Greenade (approved):

Greenade:
very nice 💯

PR #174, opened 2024-04-30
Feature/update button event info sheet

Greenade:
created a button to edit the event on the event info sheet, updated the viewmodel accordingly
image

### DIFF
--- a/app/src/main/java/com/github/swent/echo/compose/components/EventInfoSheet.kt
+++ b/app/src/main/java/com/github/swent/echo/compose/components/EventInfoSheet.kt
@@ -44,7 +44,9 @@ fun EventInfoSheet(
     event: Event,
     onJoinButtonPressed: () -> Unit,
     onDismiss: () -> Unit,
-    onFullyExtended: () -> Unit
+    onFullyExtended: () -> Unit,
+    canModifyEvent: Boolean,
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/github/swent/echo/compose/components/EventInfoSheet.kt
```diff
@@ -108,7 +110,14 @@ fun EventInfoSheet(
                 style = MaterialTheme.typography.bodyLarge
             )
 
-            Row(modifier = Modifier.align(Alignment.BottomCenter).padding(bottom = 70.dp)) {
+            val rowModifier =
```
ntissieres, line 113: can be a one-liner : Modifier.align(Alignment.BottomCenter).padding(bottom = 70.dp, end = if(canModifyEvent) 140.dp else 0.dp)

ntissieres (changes requested):
A minor change can be done to avoid unnecessary code but the rest is fine, good work !

app/src/main/java/com/github/swent/echo/compose/components/EventInfoSheet.kt
Greenade: makes sense, didn't think about that

ntissieres (approved):
Perfect !

PR #175, opened 2024-04-30
Display error message in the event creation screen

unglazedstamp:


app/src/main/java/com/github/swent/echo/compose/event/EventScreen.kt
```diff
@@ -1,52 +1,100 @@
...
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.unit.dp
 import com.github.swent.echo.R
+import com.github.swent.echo.viewmodels.event.EventStatus
 import com.github.swent.echo.viewmodels.event.EventViewModel
+import kotlinx.coroutines.launch
 
 val EVENT_PADDING_BETWEEN_INPUTS = 10.dp
 /** This screen allows the user to create or edit an event. */
+@SuppressLint("CoroutineCreationDuringComposition")
```
octogradiste, line 37: You can remove this, see below how to do it.

app/src/main/java/com/github/swent/echo/compose/event/EventScreen.kt
```diff
@@ -1,52 +1,100 @@
...
-            }
-    ) {
-        EventTitleAndBackButton(title = title, onBackButtonPressed = onEventBackButtonPressed)
-        // all the inputs for an event
-        EventPropertiesFields(eventViewModel = eventViewModel)
-        // save button
-        OutlinedButton(
+    val scope = rememberCoroutineScope()
+    val snackBarHostState by remember { mutableStateOf(SnackbarHostState()) }
+    val eventStatus = eventViewModel.status.collectAsState()
```
octogradiste, line 48: You could use `by` instead of `=`.

app/src/main/java/com/github/swent/echo/compose/event/EventScreen.kt
```diff
@@ -1,52 +1,100 @@
...
+    var saveButtonText by remember { mutableIntStateOf(R.string.edit_event_screen_save) }
+    var saveButtonClicked by remember { mutableStateOf(false) }
+    val localContext = LocalContext.current
+
+    if (saveButtonClicked && eventStatus.value is EventStatus.Saved) {
+        saveButtonClicked = false
+        saveButtonText = R.string.edit_event_screen_save
+        onEventSaved()
+    } else if (eventStatus.value is EventStatus.Error) {
+        scope.launch {
```
octogradiste, line 59: Use `LaunchEffect(eventStatus.value) {....}`

octogradiste (changes requested):
Good job! Nice addition for the UX.

app/src/main/java/com/github/swent/echo/compose/event/EventScreen.kt
unglazedstamp: Thanks I wasn't aware of this function

octogradiste (approved):

PR #176, opened 2024-04-30
📦 NEW: implement createEvent() method on repository and supabase

### LINKED ISSUE #102: implement: add create endpoints to repository
- [ ] ignores UUID which should be an empty string on creation
- [ ] returns created object (with assigned uuid)

violoncelloCH:
closes #102

### DIFF
--- a/app/src/main/java/com/github/swent/echo/data/supabase/SupabaseDataSource.kt
+++ b/app/src/main/java/com/github/swent/echo/data/supabase/SupabaseDataSource.kt
@@ -43,12 +43,27 @@ class SupabaseDataSource(private val supabase: SupabaseClient) : RemoteDataSourc
         return event.toEvent()
     }
 
+    override suspend fun createEvent(event: Event): String {
+        val eventSupabase = EventSupabaseSetter(event).copy(eventId = null)
+        val eventId =
+            supabase
+                .from("events")
+                .insert(eventSupabase) { select() }
+                .decodeSingle<EventSupabaseSetter>()
+                .eventId!!
+        setEventTagRelations(eventId, event.tags)
+        return eventId
+    }
+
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/github/swent/echo/data/supabase/SupabaseDataSource.kt
```diff
@@ -43,12 +43,27 @@ class SupabaseDataSource(private val supabase: SupabaseClient) : RemoteDataSourc
...
         val eventSupabase = EventSupabaseSetter(event)
         supabase.from("events").upsert(eventSupabase, onConflict = "event_id")
+        setEventTagRelations(event.eventId, event.tags)
+    }
 
-        val eventTags = event.tags.map { tag -> EventTagSupabase(tag.tagId, event.eventId) }
-        supabase.from("event_tags").upsert(eventTags)
+    private suspend fun setEventTagRelations(eventId: String, tags: Set<Tag>) {
+        val eventTags = tags.map { tag -> EventTagSupabase(tag.tagId, eventId) }
+        supabase.from("event_tags").upsert(eventTags, onConflict = "tag_id,event_id") { select() }
```
octogradiste, line 66: Why do we need the `select()`, couldn't we drop this?

octogradiste (approved):

app/src/main/java/com/github/swent/echo/data/supabase/SupabaseDataSource.kt
violoncelloCH: You're right, that's a leftover from debugging I've missed to remove. Here we don't want to return the just upserted row. 
I'll change this!

octogradiste (approved):
Good job, very clean code!

PR #179, opened 2024-05-01
Feature/updated HomeScreenViewModel with repository

Greenade:
updated the viewModel to use the repository. Be wary as you may not see all events as before based on the profile you log in on, for example, test@example.com is in IN and BA3 which means they won't see events made for BA6 that we could see before.

also commented 2 small unused methods in the viewModel (we might still need them later)

### DIFF
--- a/app/src/main/java/com/github/swent/echo/viewmodels/HomeScreenViewModel.kt
+++ b/app/src/main/java/com/github/swent/echo/viewmodels/HomeScreenViewModel.kt
@@ -66,11 +66,9 @@ constructor(
             )
         )
     val filtersContainer = _filtersContainer.asStateFlow()
-    private val _profileName =
-        MutableStateFlow<String>("")
+    private val _profileName = MutableStateFlow<String>("")
     val profileName = _profileName.asStateFlow()
-    private val _profileClass =
-        MutableStateFlow<String>("")
+    private val _profileClass = MutableStateFlow<String>("")
     val profileClass = _profileClass.asStateFlow()
     private var section = ""
     private var semester = ""
@@ -78,11 +76,13 @@ constructor(
     init {
         viewModelScope.launch {
             val userId = authenticationService.getCurrentUserID() ?: ""
-            allEventsList = SAMPLE_EVENTS//repository.getAllEvents()
-            allTagSet = SAMPLE_TAGS//repository.getAllTags().toSet()
+            allEventsList = SAMPLE_EVENTS // repository.getAllEvents()
+            allTagSet = SAMPLE_TAGS // repository.getAllTags().toSet()
             semester = repository.getUserProfile(userId)?.semester?.name ?: ""
             section = repository.getUserProfile(userId)?.section?.name ?: ""
-            _profileClass.value = if(semester == "") section else if (section == "") semester else "$section - $semester"
+            _profileClass.value =
+                if (semester == "") section
+                else if (section == "") semester else "$section - $semester"
             _profileName.value = repository.getUserProfile(userId)?.name ?: ""
             refreshFiltersContainer()
         }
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/github/swent/echo/viewmodels/HomeScreenViewModel.kt
```diff
@@ -66,18 +66,24 @@ constructor(
...
-    private val _profileClass =
-        MutableStateFlow<String>("IN - BA6") // placeholder until we have user profiles
+    private val _profileClass = MutableStateFlow<String>("")
     val profileClass = _profileClass.asStateFlow()
+    private var section = ""
+    private var semester = ""
 
     init {
         viewModelScope.launch {
+            val userId = authenticationService.getCurrentUserID() ?: ""
```
octogradiste, line 78: As you are in the home screen, you can safely assume that the current user id exists, e.g. `val userId = authenticationService.getCurrentUserID()!!`.

app/src/main/java/com/github/swent/echo/viewmodels/HomeScreenViewModel.kt
```diff
@@ -66,18 +66,24 @@ constructor(
...
             allEventsList = SAMPLE_EVENTS // repository.getAllEvents()
-            allTagSet = SAMPLE_TAGS // repository.getAllTags()
+            allTagSet = SAMPLE_TAGS // repository.getAllTags().toSet()
```
octogradiste, line 80: Why not use the repository to the all events and tags?

app/src/main/java/com/github/swent/echo/viewmodels/HomeScreenViewModel.kt
```diff
@@ -169,6 +175,8 @@ constructor(
     // End of methods to set the filters container values
 
     private fun filterEvents() {
+        print("section : $section\n")
+        print("semester : $semester\n")
```
octogradiste, line 179: Good for testing, but shuold be removed from main.

app/src/main/java/com/github/swent/echo/viewmodels/HomeScreenViewModel.kt
```diff
@@ -179,16 +187,16 @@ constructor(
...
+                                event.tags.any { tag ->
+                                    tag.name.lowercase() == section.lowercase()
+                                }) ||
+                            (_filtersContainer.value.classChecked &&
                                 event.tags.any { tag ->
-                                    tag.name.lowercase() == "ba6"
-                                }) // change when we have the userProfile (take their class and
-                        // section
-                        // as strings)
+                                    tag.name.lowercase() == semester.lowercase()
```
octogradiste, line 198: If the `semester` is empty (or you could also use a null value) we shuold not filter by the semester, otherwise no events will be shown. We should do the same for `section`.
Also, consider consider using multiple `.filter {}`, e.g.
```
allEventList
    .filter {} // by tags
    .filter {} // by time
    .filter {} / / by section 
    ....
```
I think it would make the code easier to read :)

octogradiste (changes requested):
Just some minor changes, good job!

app/src/main/java/com/github/swent/echo/viewmodels/HomeScreenViewModel.kt
Greenade: lol my bad I debugged something and forgot to put it back

app/src/main/java/com/github/swent/echo/viewmodels/HomeScreenViewModel.kt
Greenade: again, I debugged some stuff but forgot to remove it 🥲

Greenade:
I couldn't change the idea to be !! because it made a nullpointer Exception so I didn't change it. It shouldn't matter anyways because when the user id is "", which is the basic value I put, the userprofile returned is null, s it has the same effect.

octogradiste (approved):

PR #180, opened 2024-05-01
HomeScreenViewModel : Quick bug fix on the date filter

ntissieres:
Quick bug fix where the lowerbound was not considered if the upperbound is used. The lowerbound is now inclusive


Greenade (approved):
Looks good to me ! Thanks for the hard work 💯

PR #183, opened 2024-05-01
HamburgerMenuDrawerSheet : Removing the unnecessary stuff

ntissieres:
Juste removing the badgecount (which is not used) for hamburger menu items and removing the unnecessary highlighting of an item which remains after we clicked on it


Greenade (approved):
Everything makes sense, good job ! 🥇

Greenade (approved):
making Sonarcloud happy is important for code sustainability 🍃 💯

PR #193, opened 2024-05-01
caching/synchronisation logic

violoncelloCH:


app/src/main/java/com/github/swent/echo/data/repository/RepositoryImpl.kt
```diff
@@ -1,65 +1,209 @@
...
+                notExpired(events_last_cached_joined.getOrDefault(userId, 0), EVENT_CACHE_TTL)
+        ) {
+            return localDataSource.getJoinedEvents(userId, FETCH_ALL)
+        }
+        val nonExpired = localDataSource.getJoinedEvents(userId, EVENT_CACHE_TTL)
+        val nonExpiredIds = nonExpired.map { it.eventId }
+        val remoteUpdate = remoteDataSource.getJoinedEventsNotIn(userId, nonExpiredIds)
+        localDataSource.leaveEventsNotIn(userId, nonExpiredIds)
+        localDataSource.joinEvents(userId, remoteUpdate.map { it.eventId })
+        events_last_cached_joined.put(userId, currentTimeStamp())
```
octogradiste, line 139: Nitpicking: could be converted to an assignment `events_last_cached_joined[userId] = currentTimeStamp()`.

app/src/main/java/com/github/swent/echo/data/repository/RepositoryImpl.kt
```diff
@@ -1,65 +1,209 @@
...
+                notExpired(tags_last_cached_subs.getOrDefault(tagId, 0), TAG_CACHE_TTL)
+        ) {
+            return localDataSource.getSubTags(tagId, FETCH_ALL)
+        }
+        val nonExpired = localDataSource.getSubTags(tagId, TAG_CACHE_TTL)
+        val nonExpiredIds = nonExpired.map { it.tagId }
+        val remoteUpdate = remoteDataSource.getSubTagsNotIn(tagId, nonExpiredIds)
+        localDataSource.deleteSubTagsNotIn(tagId, nonExpiredIds)
+        localDataSource.setTags(remoteUpdate)
+        tags_last_cached_subs.put(tagId, currentTimeStamp())
```
octogradiste, line 169: Nitpicking: could also be converted to an assignment.

octogradiste (approved):
Thank you so much, massive work! The code is very verbose and therefore easy to understand. There is not much to say, every thing looks good. The only thing is that we should wrap all the calls to supabse in try catch blocks. You already did this for some of the functions but not all of them. I think it would make sense to do this everywhere, you never know when working with the network... But I think we should open an issue for this and address it in a later PR. Also, it needs some thinking on what we should do in case a call to supabase throws an exception.

app/src/main/java/com/github/swent/echo/data/repository/RepositoryImpl.kt
violoncelloCH: yes, you're right, I've considered that; however personally think calling .put is more explicit for the actual action performed, so that's why I prefer it.

app/src/main/java/com/github/swent/echo/data/repository/RepositoryImpl.kt
violoncelloCH: same as in the above comment^

violoncelloCH:
thank you for the review!
yes you're right that we should check for network related exceptions everywhere... I'll create an issue for this for next sprint

PR #195, opened 2024-05-02
fix/filters-reformed

Greenade:
new default state and more intuitive filtering (default displays everything)


octogradiste (approved):
LGTM! will make the filtering much more intuitive.

PR #197, opened 2024-05-02
Update/list mode new design

ntissieres:
New design is here ! I tried to reveal my artistic talent in order to produce the best, most beautiful, sexiest design you ever seen to display a list 🥇 
I removed all the stuff that is not implemented in the old list in order to have only functional features in the list. it should be a bit better with the new colors that i will implement just after this PR.
Don't hesitate to give your opinion, i know some visual might be discutable, like putting a different color for the epfl/section/class who appears to not be the best but i wait your feedback !


app/src/main/java/com/github/swent/echo/viewmodels/HomeScreenViewModel.kt
```diff
@@ -192,9 +193,10 @@ constructor(
                         !_filtersContainer.value.pendingChecked &&
```
srsingh04, line 193: knitpick: there are too many conditions here, personally I would do something like: 
val confirm_full_pending = !_filtersContainer.value.confirmedChecked && !_filtersContainer.value.pendingChecked &&  !_filtersContainer.value.fullChecked
wrapping the conditions in variables and then using them instead in the scope makes it easier to read.

app/src/main/java/com/github/swent/echo/compose/components/ListDrawer.kt
```diff
@@ -23,126 +28,254 @@ import androidx.compose.runtime.remember
...
                         Text(stringResource(id = R.string.list_drawer_join_event))
                     }
                 }
-
-                 */
             }
         }
     }
 }
+
```
srsingh04, line 276: You can use the @ExcludeFromJacocoGeneratedReport to exclude your preview function from the unit tests coverage repoirt and increase line coverage.

app/src/main/java/com/github/swent/echo/compose/components/ListDrawer.kt
```diff
@@ -23,126 +28,254 @@ import androidx.compose.runtime.remember
...
+    val fraction1 = 0.8f
+    val fraction2 = 0.6f
+    val listColor1 = MaterialTheme.colorScheme.primaryContainer
+    val listColor2 =
+        listColor1.copy(
+            red = listColor1.red * fraction1,
+            green = listColor1.green * fraction1,
+            blue = listColor1.blue * fraction1
+        )
+    val listColor3 =
```
srsingh04, line 80: I assume that the variations of colors is for aesthetic purposes but do not forget to comment and explain as to what their functionality is in their code.

srsingh04 (approved):
Ready to be merged.

app/src/main/java/com/github/swent/echo/compose/components/ListDrawer.kt
ntissieres: yes giving them better name would be more understandable, I will do that in the futur cleanup

app/src/main/java/com/github/swent/echo/compose/components/ListDrawer.kt
ntissieres: Nice tips ! Thanks

PR #198, opened 2024-05-02
small fix so that all events are displayed when the search is empty

Greenade:
all events are displayed when search is empty


octogradiste (approved):
LGTM! thanks for the fix!

PR #199, opened 2024-05-02
use supabase datasource

unglazedstamp:
- replace the local datasource with Supabase datasource
- fix some UI problems


violoncelloCH (approved):
nice, LGTM !
we should just make sure that the user profile creation flow is enforcing the creation of the profile before we navigate to the map before merging this

PR #200, opened 2024-05-02
Update/new color scheme

ntissieres:
New color scheme plus minor fixes to fit it


octogradiste (approved):
Thanks for the new colors! LGTM

PR #201, opened 2024-05-02
Integrate modify event screen

octogradiste:
I had to change how the navigation works to allow navigating with some arguments.


Greenade (approved):
Looks clean ! With this I think the epics are quite complete ! 💯

PR #202, opened 2024-05-02
Add end to end tests

octogradiste:
Adds end-to-end tests for both of our user flows.

### DIFF
--- a/app/src/androidTest/java/com/github/swent/echo/MainActivityTest.kt
+++ b/app/src/androidTest/java/com/github/swent/echo/MainActivityTest.kt
@@ -173,22 +173,15 @@ class MainActivityTest {
         // Registers with google sign in
         composeTestRule.onNodeWithTag("google-sign-in-button").performClick()
 
-        // TODO: Use create profile page once we have it
-        runBlocking {
-            val userId = authenticationService.getCurrentUserID()
-            assertNotNull(userId)
-            repository.setUserProfile(
-                UserProfile(
-                    userId!!,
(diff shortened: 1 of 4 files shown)


octogradiste:
We still need to integrate the user profile creation page in our end-to-end tests. Waiting for #188 to be merged.

octogradiste:
Note that this will close #150.

ntissieres (dismissed):
The user flows are perfectly executed and tested, thank you for the nice work !

app/src/androidTest/java/com/github/swent/echo/MainActivityTest.kt
```diff
@@ -79,4 +106,147 @@ class MainActivityTest {
...
+
+        // Fill the event creation form
+        val title = "This is a test event title"
+        val description = "This is a test event description"
+        composeTestRule.onNodeWithTag("Title-field").performTextInput(title)
+        composeTestRule.onNodeWithTag("Description-field").performTextInput(description)
+        composeTestRule.onNodeWithTag("nb-participant-field").performTextReplacement("10")
+
+        // Click outside of the text field
+        composeTestRule.onNodeWithTag("Title-field").requestFocus()
```
violoncelloCH, line 203: why is this performed here ? does event creation fail otherwise ?

violoncelloCH (approved):
Nice! Only thing that comes to mind is that we could actually test setting more / all fields in the event creation/edit. But that can go to a new task I think :)

PR #210, opened 2024-05-02
Ensure user has profile

octogradiste:
When navigating to the home screen, the `NavigationActions` checks if the user is logged in and has a profile.

### DIFF
--- a/app/src/main/java/com/github/swent/echo/viewmodels/authentication/CreateProfileViewModel.kt
+++ b/app/src/main/java/com/github/swent/echo/viewmodels/authentication/CreateProfileViewModel.kt
@@ -1,7 +1,6 @@
 package com.github.swent.echo.viewmodels.authentication
 
 import androidx.lifecycle.ViewModel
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/github/swent/echo/viewmodels/authentication/CreateProfileViewModel.kt
```diff
@@ -61,15 +60,17 @@ constructor(
...
-                // getUserProfile function so that it returns null if a user has not yet created a
-
+        // viewModelScope.launch {
+        val userId = authenticationService.getCurrentUserID()
+        if (userId == null) {
+            _errorMessage.value = "Profile creation error: Not logged in"
+            // TODO: handle error later on. Will this ever be the case?
+        } else {
+            // TODO: this is the logic for creating a userProfile, changes are required in the
+            // getUserProfile function so that it returns null if a user has not yet created a
```
violoncelloCH, line 70: I think this is done with the last commit

PR #236, opened 2024-05-04
Local data source `syncedSecondsAgo` implementation

octogradiste:


octogradiste:
The last 8 times I ran the CI, all tests passed. I think it failed once because the emulator crashed because of some other reason... I added the `--info` flag for the tests so in the future we can see what caused the crash...

violoncelloCH (approved):
Looks all good to me. Thanks for the work!

PR #237, opened 2024-05-05
fix wrong number of participant in edit event screen

unglazedstamp:


srsingh04 (approved):
Code ready to be merged, thank you for the fix!

PR #240, opened 2024-05-06
Event Creation Map

### LINKED ISSUE #221: Make the location selection composable for event creation
make the composable/screen to select the location for the event creation screen

alejandrocalles:
Should be able to move a pin around to select a location.

Closes #221.


app/src/main/java/com/github/swent/echo/compose/map/LocationSelector.kt
```diff
@@ -0,0 +1,56 @@
...
+        Circle(
+            center = markerCenter,
+            radius = 30F,
+            isDraggable = true,
+            color = "Red",
+            opacity = 0.1F,
+            zIndex = 1,
+            onCenterDragged = { markerCenter = it },
+            onDragFinished = {
+                mapCenter = it
```
violoncelloCH, line 44: Making the map center every time one dragged the marker also seems to make it zoom in on that point. This is not really useful since one most likely has to zoom out again to actually get closer to the exact location one wanted to select. Disabling the centering here would make the location selector easier to use.

violoncelloCH (changes requested):
Thanks for working on this. Well done!
I have two suggestions below as discussed.
Further on, but probably rather for a follow up since it will need quite some restructuring, I'd suggest to make the marker actually stick to the center of the screen/selector dialog. The user then simple moves the map below the marker to choose the location. This way only one thing needs to be moved instead of two, making the process easier. For illustrative purposes, a quick example video showing how this would work:

https://github.com/echo-swent/echo/assets/22591354/2de42d78-e287-49db-b1b6-284a171c1edb

app/src/main/java/com/github/swent/echo/compose/map/LocationSelector.kt
```diff
@@ -0,0 +1,56 @@
...
+    val context = LocalContext.current
+    val styleUrl =
+        context.getString(R.string.maptiler_base_style_url) +
+            context.getString(R.string.maptiler_api_key)
+    var markerCenter by remember { mutableStateOf(initialLocation) }
+    var mapCenter by remember { mutableStateOf(initialLocation) }
+    MapLibre(
+        modifier = modifier.testTag("location-selector-map-libre"),
+        styleUrl = styleUrl,
+        cameraPosition = CameraPosition(target = mapCenter, zoom = DEFAULT_ZOOM)
```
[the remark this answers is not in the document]
alejandrocalles: The way the event screen currently works is that the values of the `Event` object being edited are used. This means that to have a specific "default" location, the `EMPTY` object for the `Location` data class should probably use `MAP_CENTER` instead of `(0.0, 0.0)`. 
I'm not sure how we would go about using the GPS. In my mind, the way events should be created is by selecting the location first, which then would be passed to the current event creation screen, but that would need a fair amount of work.

app/src/androidTest/java/com/github/swent/echo/compose/map/LocationSelectorTest.kt
```diff
@@ -0,0 +1,27 @@
...
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+class LocationSelectorTest {
+
+    @get:Rule val composeTestRule = createComposeRule()
+
+    @Test
+    fun shouldDisplayLocationSelector() {
+        composeTestRule.setContent { LocationSelector {} }
```
octogradiste, line 18: Nitpicking: this could be done in setUp function annotated with @Before

app/src/main/java/com/github/swent/echo/compose/event/EventLocationEntry.kt
```diff
@@ -1,115 +1,97 @@
...
-                Modifier.padding(horizontal = 5.dp, vertical = 20.dp)
-                    .align(Alignment.Bottom)
-                    .testTag("Location-button"),
-            onClick = { showPickLocationDialog = true }
-        ) {
-            Icon(
-                imageVector = Icons.Outlined.Place,
-                contentDescription =
-                    stringResource(R.string.edit_event_screen_select_location_button)
+        Box(modifier = Modifier.aspectRatio(1.5F)) {
```
octogradiste, line 53: To be consistent with the rest of the form, I'd suggest to add a `.padding(EVENT_PADDING_BETWEEN_INPUTS)`. Otherwise, the map is not aligned with the other input fields...

app/src/main/java/com/github/swent/echo/compose/event/EventLocationEntry.kt
```diff
@@ -1,115 +1,97 @@
...
+            onLocationChanged(it)
+        }
     }
-    Row(modifier = Modifier.fillMaxWidth(1f)) {
+    Column(modifier = Modifier.fillMaxWidth(1f)) {
         EventTextEntry(
             name = stringResource(R.string.edit_event_screen_location),
             value = location.name,
             modifier = Modifier.fillMaxWidth(0.8f)
         ) {
```
octogradiste, line 50: Here I'd remove the `0.8f` as there is no more a marker icon to open the location dialog. Makes it more consistent with the rest...

app/src/main/java/com/github/swent/echo/compose/event/EventLocationEntry.kt
```diff
@@ -1,115 +1,97 @@
...
+            onDismissRequest = { editLocation = false }
+        ) {
+            displayedLocation = it
+            onLocationChanged(it)
+        }
     }
-    Row(modifier = Modifier.fillMaxWidth(1f)) {
+    Column(modifier = Modifier.fillMaxWidth(1f)) {
         EventTextEntry(
             name = stringResource(R.string.edit_event_screen_location),
```
octogradiste, line 47: We could change this to "Location Title" (addressing what the coaches pointed out)

app/src/main/java/com/github/swent/echo/compose/event/EventLocationEntry.kt
```diff
@@ -1,115 +1,97 @@
...
-                    stringResource(R.string.edit_event_screen_select_location_button)
+        Box(modifier = Modifier.aspectRatio(1.5F)) {
+            LocationDisplayer(position = displayedLocation.toLatLng())
+            Box(
+                modifier =
+                    Modifier.matchParentSize()
+                        .clickable(
+                            interactionSource = remember { MutableInteractionSource() },
+                            indication = null,
+                            onClick = {}
```
octogradiste, line 61: We could use `onClick = { editLocation = true)` to open the dialog when the user clicks on the map. I feel like this is quite intuitive. What do you think?

app/src/main/java/com/github/swent/echo/compose/event/EventLocationEntry.kt
```diff
@@ -1,115 +1,97 @@
...
+                        )
             )
         }
+        Button(modifier = Modifier.testTag("Location-button"), onClick = { editLocation = true }) {
+            Text("Edit")
+        }
```
octogradiste, line 67: If we to the above, we can remove this button.

octogradiste (changes requested):
Thanks a lot for your work! I pointed out just some minor things to imporve the UX. Othwersise, everything is very clear and well coded!

app/src/androidTest/java/com/github/swent/echo/compose/map/LocationSelectorTest.kt
alejandrocalles: Since the content for both tests is different, I would think it's cleaner to leave it in the test itself. Otherwise I'm not sure how to set both contents simultaneously.

app/src/main/java/com/github/swent/echo/compose/event/EventLocationEntry.kt
alejandrocalles: I don't remember that well, but I think @unglazedstamp asked me to leave this to him. I've opened an issue: #273.

app/src/main/java/com/github/swent/echo/compose/event/EventLocationEntry.kt
alejandrocalles: You're right, thanks for catching this :)

app/src/androidTest/java/com/github/swent/echo/compose/map/LocationSelectorTest.kt
octogradiste: Yeah, you're right, I didn't read it carefully enough. In this case, it makes no sense to add a setUp function!

octogradiste (approved):

PR #243, opened 2024-05-07
Feature/association list screen

unglazedstamp:
Add two screen to see the association memberships and the followed associations of the user

### DIFF
--- a/app/src/main/java/com/github/swent/echo/compose/association/AssociationListScreen.kt
+++ b/app/src/main/java/com/github/swent/echo/compose/association/AssociationListScreen.kt
@@ -3,7 +3,6 @@ package com.github.swent.echo.compose.association
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.padding
-import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.material3.Icon
 import androidx.compose.material3.ListItem
--- a/app/src/main/java/com/github/swent/echo/compose/association/AssociationListScreen.kt
+++ b/app/src/main/java/com/github/swent/echo/compose/association/AssociationListScreen.kt
@@ -4,6 +4,7 @@ import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.lazy.LazyColumn
+import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material3.Icon
 import androidx.compose.material3.ListItem
 import androidx.compose.material3.MaterialTheme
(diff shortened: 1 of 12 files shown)


app/src/main/java/com/github/swent/echo/compose/association/AssociationListScreen.kt
```diff
@@ -0,0 +1,106 @@
...
+fun AssociationListElement(
+    association: Association,
+    hasActionButton: Boolean,
+    actionButtonName: String,
+    onActionButtonClicked: () -> Unit
+) {
+    var expanded by remember { mutableStateOf(false) }
+    ListItem(
+        modifier =
+            Modifier.padding(vertical = 7.dp, horizontal = 15.dp).clickable {
```
ntissieres, line 63: I think it could be good if you use clip(RoundedCornerShape(5.dp)) to keep the same look for the lists over the app

app/src/main/java/com/github/swent/echo/compose/association/AssociationListScreen.kt
```diff
@@ -0,0 +1,106 @@
...
+        overlineContent = { Spacer(modifier = Modifier) }, // used to align the icon and the button
+        headlineContent = {
+            Text(
+                text = association.name,
+                modifier = Modifier.testTag("${association.associationId}-name"),
+                style = MaterialTheme.typography.titleMedium
+            )
+        },
+        supportingContent = {
+            Text(
```
ntissieres, line 75: I'm not sure about the necessity to show the description of the association here. In fact, the user already followed the association so she/he knows what the association is about. And it simplifies the look.

ntissieres (changes requested):
The code is good, but maybe two corrections : 
- Add rounded corner shape to the list items
- Remove the description which is not really necessary imo

app/src/main/java/com/github/swent/echo/compose/association/AssociationListScreen.kt
unglazedstamp: yes, thanks for the suggestion

app/src/main/java/com/github/swent/echo/compose/association/AssociationListScreen.kt
unglazedstamp: That's a good point, but I believe that's not necessarily true for a large number of associations. I think it makes sense to remove the description in the committee association list screen, but not in the followed associations list screen as it will be used to unfollow associations which are no more interesting for the user. In this case it can help to be able to easily look at the description to remember what an association is about.

violoncelloCH (dismissed):
Thanks for the work on this!
Sorry to come up with this only now, but as mentioned over at https://github.com/echo-swent/echo/pull/251#discussion_r1594763526 , I don't think it makes sense to have unfollow buttons this prominently in the list. Please read my linked comment and reconsider if we really want to have those buttons.
If we decide on keeping them, there should at east have an easy way to undo the action. I.e. the association should stay in the list, showing an "follow" button at the place of the "unfollow" so a user can easily revert their accidential action.

app/src/main/java/com/github/swent/echo/compose/association/AssociationListScreen.kt
unglazedstamp: After discussion in the meeting I understand better the what's the purpose of these screen, so I removed the description.

unglazedstamp:
Thank you for the suggestion, I still think that it's useful to have the follow button. I agree that the user should be able to undo his action, so I added a snackbar with an undo button which appears when the user unfollow an association.

ntissieres (dismissed):
The changes are okay for me ! Nice work ;)

ntissieres (approved):

PR #245, opened 2024-05-07
Fix/adaptative display scope event

ntissieres:
in the list mode, use the viewModel to display the right semester and section


Greenade (approved):
based on the discussion we had on telegram, this makes perfect sense. Thank you for your work this looks perfect to me !

Issue #190, opened 2024-05-01
state object which holds the current network state, for composables and repository


violoncelloCH:
current connectivity state: https://developer.android.com/develop/connectivity/network-ops/reading-network-state#instantaneous
callback registration on state change: https://developer.android.com/training/monitoring-device-state/connectivity-status-type
the capability we want to check: https://developer.android.com/reference/android/net/NetworkCapabilities#NET_CAPABILITY_INTERNET

octogradiste:
Thanks for adding those links! I feel like we should check for the [NET_CAPABILITY_VALIDATED](https://developer.android.com/reference/android/net/NetworkCapabilities#NET_CAPABILITY_VALIDATED). As `NET_CAPABILITY_INTERNET` doesn't mean you can reach public servers...

Also, I've been thinking on how to implement this. I suggest something like this:

```kotlin
interface NetworkService {
    // Useful for the UI
    val isOnline: StateFlow<Boolean>

    // Useful for the Repository
    fun isOnline(): Boolean
}
```

It would be really easy to configure this as a singleton which we could inject with hilt in any part of our code which needs this information.

unglazedstamp:
Good idea, you could rename the function `isOnlineNow()` to differentiate the function from the stateflow.

PR #246, opened 2024-05-07
offline mode_new_branch

srsingh04:
Tried to push a clean branch

### DIFF
--- a/app/src/main/java/com/github/swent/echo/compose/components/HomeScreen.kt
+++ b/app/src/main/java/com/github/swent/echo/compose/components/HomeScreen.kt
@@ -60,6 +60,7 @@ fun HomeScreen(navActions: NavigationActions, homeScreenViewModel: HomeScreenVie
             Content(paddingValues, navActions, homeScreenViewModel)
         }
     }
+    ConnectivityStatus()
 }
 
 @Composable
--- a/app/src/main/java/com/github/swent/echo/compose/components/OfflineUI.kt
+++ b/app/src/main/java/com/github/swent/echo/compose/components/OfflineUI.kt
@@ -1,0 +1,100 @@
+package com.github.swent.echo.compose.components
+
+import androidx.compose.animation.AnimatedVisibility
+import androidx.compose.animation.animateColorAsState
+import androidx.compose.animation.expandHorizontally
+import androidx.compose.animation.shrinkHorizontally
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.PaddingValues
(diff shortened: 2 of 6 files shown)


app/src/main/java/com/github/swent/echo/compose/components/HomeScreen.kt
```diff
@@ -60,6 +60,7 @@ fun HomeScreen(navActions: NavigationActions, homeScreenViewModel: HomeScreenVie
             Content(paddingValues, navActions, homeScreenViewModel)
         }
     }
+    ConnectivityStatus()
```
octogradiste, line 63: We can remove this for now, we will add it back to the content of the home screen once we have the `NetworkService` implemented.

app/src/main/java/com/github/swent/echo/compose/components/OfflineUI.kt
```diff
@@ -0,0 +1,104 @@
...
+import com.github.swent.echo.connectivity.connectivityState
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.delay
+
+@ExperimentalCoroutinesApi
+@Composable
+fun ConnectivityStatus() {
+    Log.d("ConnectivityStatus", "ConnectivityStatus started")
+    // This will cause re-composition on every network state change
+    val connection by connectivityState()
```
octogradiste, line 40: Remove `val connection by connectivityState()`, instead add a `connection` parameter. It makes it more modular and will integrate well with the `NetworkService`.

octogradiste (changes requested):
Thanks a lot for working on this! As discussed today after the meeting, we are planning to add a `NetworkService` (#190) which will be used in the whole app.

app/src/main/java/com/github/swent/echo/compose/navigation/AppNavigationHost.kt
```diff
@@ -35,8 +35,8 @@ fun AppNavigationHost(
 
     NavHost(
         navController = navController,
-        // startDestination = Routes.PROFILE_CREATION.name,
-        startDestination = if (userIsLoggedIn) Routes.MAP.name else Routes.REGISTER.name,
+        startDestination = Routes.PROFILE_CREATION.name,
+        // startDestination = if (userIsLoggedIn) Routes.MAP.name else Routes.REGISTER.name,
```
octogradiste, line 39: If you uncomment this line, the end-to-end tests should pass again ;)

octogradiste (changes requested):
Thanks for addressing the requested changes, but as changed the startDestination of the app, the end-to-end tests fail...

Issue #192, opened 2024-05-01
allow to subscribe/join to Events


violoncelloCH:
- [x] add relation in supabase
- [x] add a setter (and deleter) on the `RemoteDataSource` and figure out how to send the query to supabase
- [x] add relation to room
- [x] add a setter (and deleter) to `LocalDataSource`
- [x] add setter on `Repository`
- [ ] connect the button in the UI to it

octogradiste:
I think this is something which the user can only do when he is online. Otherwise, we would have to synchronize when he is back online. 

I'll still implement this functionality in room, but only for caching, the repository should ensure that nothing is modified while the app is offline.

PR #249, opened 2024-05-08
add ability to join, leave and get joined events for user to repository and supabase

violoncelloCH:
works towards closing #192

I've also added corresponding RLS policies on Supabase as well as a Postgrest function as well as triggers to increment/decrement the counter of participants on the corresponding `Event`

### DIFF
--- a/app/src/test/java/com/github/swent/echo/data/repository/RepositoryImplTest.kt
+++ b/app/src/test/java/com/github/swent/echo/data/repository/RepositoryImplTest.kt
@@ -90,6 +90,32 @@ class RepositoryImplTest {
     }
 
     @Test
+    fun joinEventTest() {
+        every {
+            runBlocking { mockedRemoteDataSource.joinEvent(userProfile.userId, event.eventId) }
+        } returns true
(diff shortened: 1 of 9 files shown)


app/src/test/java/com/github/swent/echo/data/repository/RepositoryImplTest.kt
```diff
@@ -89,6 +89,32 @@ class RepositoryImplTest {
         assertEquals(Arrays.asList(event), eventsResult)
     }
 
+    @Test
+    fun joinEventTest() {
+        every {
```
octogradiste, line 94: Tip: you could use `coEvery{}` and remove the `runBlocking`.

octogradiste (approved):
Looks good to me! Very easy to read and understand, thanks!

app/src/test/java/com/github/swent/echo/data/repository/RepositoryImplTest.kt
violoncelloCH: thanks for the hint! I'll use that next time :)

PR #250, opened 2024-05-08
linked the TagUI to the homescreen, adapted the viewmodel

Greenade:
used the tagUI in the homescreen and filtered events based on that. Two modes are possible, filtering based on the followed tags, or filtering based on the search sheet.


ntissieres (changes requested):
Good code, maybe 2 littles changes : 
- When you click on an already selected tag, it might be good to unselect it. Because actually, it's not possible to go back and select no category.
- It might be good to have the ListDrawer to be drawn under the TagsUI

app/src/main/java/com/github/swent/echo/compose/components/HomeScreen.kt
```diff
@@ -90,7 +93,13 @@ private fun Content(
                 callback = homeScreenViewModel::onEventSelected,
             )
         }
-        // add the tag filtering here
+
```
ntissieres, line 96: at line 87, replace by :
Box(modifier = Modifier.padding(paddingValues)) {
    if (mode == MapOrListMode.LIST) {
        Column(){
            if(tags.isNotEmpty()){
                TagUI(...)
            }
            ListDrawer(...)
        }else{
            TagUI(...)
            MapDrawer(...)
        }
    }
}

ntissieres (approved):
Good changes ;) Thanks for your work !

PR #255, opened 2024-05-09
better search bar

Greenade:
we can now search in the event title and in the event description


ntissieres (approved):
Nice update, thanks for your work !

PR #256, opened 2024-05-09
make repository return nullable objects

violoncelloCH:
This makes the repository return null, if the requested object does not exist.
For now I simply forced the type in the viewmodels: @unglazedstamp could you please modify those occurrences cleanly? You'll find the occurrences in the following commit: ad26b677f3fcedeb3f2a96990cb1a89cad818e69. You can directly push a commit to this branch so we can merge this together once it's ready :)


octogradiste (approved):
This really had to be changed, thanks for doing the work! Everything looks good to me!

PR #269, opened 2024-05-10
Fix null user profile bug

### LINKED ISSUE #105: Create a loading screen to initialize the `AuthenticationService`

### LINKED ISSUE #258: Redirect to the user profile creation page at application start when the user has not profile yet

octogradiste:
This will close #258, the bug the coaches pointed out during the last scrum retrospective.

To solve it, I moved the logic from the `NavigationActions` to the `AppNavHost`. This allows us to do the checks at application start. I added a loading animation to avoid blocking the app during the checks. This will close #105.

Note that as the authentication relied on the `NavigationActions` for navigating to the correct route (profile creation or map), there were some tests which broke. So I had to put some more logic into the authentication view models.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -130,6 +130,7 @@ dependencies {
     testImplementation(libs.slf4j.simple)
     testImplementation(libs.robolectric)
     testImplementation(libs.androidx.ui.test.junit4)
+    testImplementation(libs.turbine)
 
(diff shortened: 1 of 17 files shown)


ntissieres (approved):
I try this version on my phone, and I confirm that it solves the bug. Thanks for your hard work and good code !

PR #271, opened 2024-05-11
Implement a network service

### LINKED ISSUE #190: state object which holds the current network state, for composables and repository

octogradiste:
This closes #190 and #270.

### DIFF
--- a/app/src/main/java/com/github/swent/echo/connectivity/NetworkServiceImpl.kt
+++ b/app/src/main/java/com/github/swent/echo/connectivity/NetworkServiceImpl.kt
@@ -1,0 +1,40 @@
+package com.github.swent.echo.connectivity
+
+import android.net.ConnectivityManager
+import android.net.Network
+import android.net.NetworkCapabilities
+import android.net.NetworkRequest
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
+
+class NetworkServiceImpl(connectivityManager: ConnectivityManager) : NetworkService {
+    private val _isOnline = MutableStateFlow(false)
+
+    init {
+        val caps = connectivityManager.getNetworkCapabilities(connectivityManager.activeNetwork)
+        _isOnline.value = caps?.hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET) ?: false
+
+        val request =
+            NetworkRequest.Builder()
+                .addCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET)
+                .build()
+
+        val networkCallback =
+            object : ConnectivityManager.NetworkCallback() {
+                override fun onAvailable(network: Network) {
+                    _isOnline.value = true
+                }
+
+                override fun onLost(network: Network) {
+                    _isOnline.value = false
+                }
+            }
+
+        connectivityManager.registerNetworkCallback(request, networkCallback)
+    }
+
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/github/swent/echo/connectivity/NetworkServiceImpl.kt
```diff
@@ -0,0 +1,40 @@
...
+
+        val request =
+            NetworkRequest.Builder()
+                .addCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET)
+                .build()
+
+        val networkCallback =
+            object : ConnectivityManager.NetworkCallback() {
+                override fun onAvailable(network: Network) {
+                    _isOnline.value = true
```
srsingh04, line 26: Knitpick: What do you think about updating the _isOnline.value asynchronously in the main thread using CoroutineScope(Dispatchers.Main).launch { _isOnline.value = true }? Doing this would prompt the UI quickly as soon as the change in network has been detected. What do you think?

app/src/test/java/com/github/swent/echo/connectivity/NetworkServiceImplTest.kt
```diff
@@ -0,0 +1,98 @@
...
+
+    @Before
+    fun setUp() {
+        connectivityManager = mockk()
+    }
+
+    @Test
+    fun `should not be online when no network`() {
+        every { connectivityManager.activeNetwork } returns mockk()
+        every { connectivityManager.getNetworkCapabilities(any()) } returns
```
srsingh04, line 26: Knitpick: The following lines of code are duplicated in each test method: 
every { connectivityManager.activeNetwork } returns mockk()
        every { connectivityManager.getNetworkCapabilities(any()) } returns
            mockk { every { hasCapability(any()) } returns true }

        every {
            connectivityManager.registerNetworkCallback(
                any<NetworkRequest>(),
                any<ConnectivityManager.NetworkCallback>()
            )
        } returns Unit
Consider wrapping them in a function (part of it could be done in @Before function setup) in order to avoid code duplication.

srsingh04 (approved):
Nice work! I have left some comments to be taken into consideration in near future for good code practices ;)
Ready to be merged.

app/src/main/java/com/github/swent/echo/connectivity/NetworkServiceImpl.kt
octogradiste: As this is a very in-expensive operation (just set a value) I don't think we need to launch a coroutine for that. As launching a coroutine requires some work, idk if it really would be quicker...

app/src/test/java/com/github/swent/echo/connectivity/NetworkServiceImplTest.kt
octogradiste: Yeah, you're right, I'll move it into the `@Before`!

octogradiste: Actually, I couldn't use the `@Before` annotation, as I need to pass an argument to the `setUp` function. But it still makes the code much cleaner, thanks!

PR #277, opened 2024-05-13
add support for offline mode in the create/edit event screen & fixes

unglazedstamp:
- add support for offline mode in the create/edit event screen
- fix wrong log
- rename location text field

### DIFF
--- a/app/src/main/java/com/github/swent/echo/compose/event/EventScreen.kt
+++ b/app/src/main/java/com/github/swent/echo/compose/event/EventScreen.kt
@@ -52,6 +52,8 @@ fun EventScreen(
     var saveButtonClicked by remember { mutableStateOf(false) }
     val localContext = LocalContext.current
 
+    val isOnline by eventViewModel.isOnline.collectAsState()
+
     if (saveButtonClicked && eventStatus is EventStatus.Saved) {
         saveButtonClicked = false
         saveButtonText = R.string.edit_event_screen_save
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/github/swent/echo/compose/event/EventScreen.kt
```diff
@@ -150,7 +154,8 @@ fun EventPropertiesFields(eventViewModel: EventViewModel) {
         }
         EventDropDownSelectOrganizer(
             organizerName = organizerName,
-            organizerList = organizerListState.value
+            organizerList = organizerListState.value,
+            enabled = isOnline
         ) {
             eventViewModel.setOrganizer(it)
         }
```
alejandrocalles, line 161: It might be worth it to pass the `isOnline` parameter to the `EventLocationEntry` so that the map can be hidden, or at least so that the location edition can be disabled.

app/src/main/java/com/github/swent/echo/viewmodels/event/EventViewModel.kt
```diff
@@ -35,6 +37,7 @@ constructor(
     val isEventNew = _isEventNew.asStateFlow()
     private val _organizerList = MutableStateFlow<List<String>>(listOf())
     val organizerList = _organizerList.asStateFlow()
+    val isOnline = networkService.isOnline
```
alejandrocalles, line 40: I'm not sure it makes sense to get the `isOnline` parameter only once. Could it be a problem if the device goes offline after this state is read?

alejandrocalles (changes requested):
Very clean code, thanks a lot! I left a couple of comments, but we can create issues for the next sprint if necessary.

app/src/main/java/com/github/swent/echo/compose/event/EventScreen.kt
unglazedstamp: Thanks for the suggestion

app/src/main/java/com/github/swent/echo/viewmodels/event/EventViewModel.kt
unglazedstamp: The networkService returns a StateFlow, so the composable will be updated when the value changes

alejandrocalles (approved):

alejandrocalles:
All good, thanks :)

PR #279, opened 2024-05-13
search mode interactions with UI

Greenade:
The follow tags UI doesn't appear when in search mode, a new button on the topbar appears in search mode to exit the mode. Status checkboxes are now exclusive, meaning only one can be selected at a time.


app/src/main/java/com/github/swent/echo/compose/components/TopAppBar.kt
```diff
@@ -55,19 +57,28 @@ fun TopAppBar(
...
                 Icon(
                     painter =
                         if (mode == MapOrListMode.MAP) {
                             rememberVectorPainter(image = Icons.Filled.List)
                         } else {
                             painterResource(id = R.drawable.map_icon)
                         },
-                    contentDescription = "Search icon to access the search screen"
+                    contentDescription =
+                        "list/map mode switch icon to switch between list and map mode"
```
srsingh04, line 81: +ve point: contentDescription made more apt so that it's easier to what its function is.

srsingh04 (approved):
Thank you for making the changes! The quality of the code is also better than its previous version, so well done! 
Ready to be merged to the main.

PR #281, opened 2024-05-14
offline mode disables the buttons in the homescreen

Greenade:
The join event and modify event buttons are now disabled when offline, both in event info sheet and list mode


ntissieres (approved):
The code is good and tests on my phone concluent. Thanks for your work !

PR #283, opened 2024-05-14
Overwrite back button behavior

### LINKED ISSUE #280: Overwrite hardware back button in situations which do not make sense
Screens affected:

- `LoginScreeen`
- `RegisterScreen`
- `LoadingScreen`
- `ProfileCreation`
- `HomeScreen`

In all of those cases, the only thing to do is to quit the app. E.g. returning from the `HomeScreen` to the `LoadingScreen` makes no sense, as it will simply load forever.

octogradiste:
Closes #280.

### DIFF
--- a/app/src/main/java/com/github/swent/echo/compose/navigation/AppNavigationHost.kt
+++ b/app/src/main/java/com/github/swent/echo/compose/navigation/AppNavigationHost.kt
@@ -1,8 +1,11 @@
 package com.github.swent.echo.compose.navigation
 
 // import com.github.swent.echo.compose.authentication.ProfileCreationScreen
+import android.app.Activity
+import androidx.activity.compose.BackHandler
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/github/swent/echo/compose/navigation/AppNavigationHost.kt
```diff
@@ -110,3 +119,10 @@ fun AppNavigationHost(
         }
     }
 }
+
+/** Quit the app when the hardware back button is pressed */
+@Composable
+fun QuitAppOnHardwareBackButtonPressPress() {
```
ntissieres, line 125: I don't know if it's a small typo but you wrote 2 times "...PressPress" in the function name

ntissieres (approved):
The app seems to crash when we try to register (with a new e-mail or an already used one), but apprently it was already the case on previous version. But considering the code it shouldn't be because of this PR. I wasn't able to test the case when you are in profile creation page, but since the code is relatively simple it should also work ! Thanks for your work.

PR #287, opened 2024-05-14
Feature/my events page

ntissieres:
Implementation of the my events page. It displays the events joined by the user and the events created by the user. For now, it doesn't include the view model so it takes fictive values for the events, but it was written to simply include it once it's done.


srsingh04 (approved):
Thank you for working on the composable. Ready to be merged!

PR #289, opened 2024-05-15
Fix app crash when register with an existing account

octogradiste:
The crash was due to the fact that the registering was successful, but didn't save the `userId` locally. To do so, we must request a login from the user. Only then the `userId` will be saved.

### DIFF
--- a/app/src/main/java/com/github/swent/echo/viewmodels/authentication/AbstractAuthenticationViewModel.kt
+++ b/app/src/main/java/com/github/swent/echo/viewmodels/authentication/AbstractAuthenticationViewModel.kt
@@ -57,18 +57,17 @@ abstract class AbstractAuthenticationViewModel : ViewModel() {
         }
     }
 
-    /**
-     * Returns the next route depending on the value of the user profile. If the profile is null, it
-     * returns [Routes.PROFILE_CREATION] and [Routes.MAP] otherwise. Note you should only call this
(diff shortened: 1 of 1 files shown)


ntissieres (changes requested):
We still can try to register with an email already used on an existing account. We are redirected in the login screen with the message "confirm your email to continue". It would be better to add an error case "email already used".
Then, if we create a new account and reach the profile creation, clicking on the back button will just reload the profile creation screen. It would be better to be redirected to the login screen.
But good news, the app doesn't crash anymore ! So it just need the fixes mentionned above.

octogradiste:
>  We are redirected in the login screen with the message "confirm your email to continue". It would be better to add an error case "email already used". 

I'll have to investigate that. I know it's confusing to the user, but I'm not sure if I can get this information somewhere. As this is only a rare case (user tries to register with an existing account) I thought this solution was acceptable.

> Then, if we create a new account and reach the profile creation, clicking on the back button will just reload the profile creation screen.

This is something we should address in another PR. Actually, this shouldn't be very hard, change the back button to log the user out and navigate to the login screen. Or we simply drop the back button... why would anyone want to go back during the profile creation? Nevertheless, we should open an issue and create another PR.

ntissieres:
Okay, it's fine for me 👍

ntissieres (approved):

PR #291, opened 2024-05-15
Feature/association viewmodel

ntissieres:
Creating the viewModel for the association

### DIFF
--- a/app/src/main/java/com/github/swent/echo/viewmodels/association/AssociationViewModel.kt
+++ b/app/src/main/java/com/github/swent/echo/viewmodels/association/AssociationViewModel.kt
@@ -1,0 +1,142 @@
+package com.github.swent.echo.viewmodels.association
+
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.viewModelScope
+import com.github.swent.echo.authentication.AuthenticationService
+import com.github.swent.echo.connectivity.NetworkService
+import com.github.swent.echo.data.model.Association
+import com.github.swent.echo.data.model.Event
+import com.github.swent.echo.data.repository.Repository
+import dagger.hilt.android.lifecycle.HiltViewModel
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/github/swent/echo/viewmodels/association/AssociationViewModel.kt
```diff
@@ -0,0 +1,142 @@
...
+import com.github.swent.echo.data.model.Association
+import com.github.swent.echo.data.model.Event
+import com.github.swent.echo.data.repository.Repository
+import dagger.hilt.android.lifecycle.HiltViewModel
+import javax.inject.Inject
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.launch
+
+enum class AssociationPage(var association: Association) {
```
Greenade, line 16: why do you need an empty association for every screen ?

ntissieres: I agree it's a bit weird but i didn't find a better solution. It's for DETAILS which need to take an Association as argument to know which association we want to display details

Greenade (changes requested):
Nice work ! Not sure why you need en empty associations for every screen though ? Everything else looks good to me :)

app/src/main/java/com/github/swent/echo/viewmodels/association/AssociationViewModel.kt
Greenade: Oh okay, I don't know if there is a better solution either 🤔 It's okay for now and we could always change it later

Greenade (approved):

Greenade:
according to the discussion above, I removed the requested changes

PR #292, opened 2024-05-15
Feature/association screen

ntissieres:
Adding the managing screen which will decide which page to display between AssociationMainScreen, AssociationDetails and AssociationSearch

### DIFF
--- a/app/src/main/java/com/github/swent/echo/compose/association/AssociationDetails.kt
+++ b/app/src/main/java/com/github/swent/echo/compose/association/AssociationDetails.kt
@@ -1,6 +1,9 @@
 package com.github.swent.echo.compose.association
 
+import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.testTag
 import com.github.swent.echo.data.model.Association
 import com.github.swent.echo.data.model.Event
 
@@ -10,4 +13,6 @@ fun AssociationDetails(
     association: Association,
     isFollowed: Boolean,
     events: List<Event>
-) {}
+) {
+    Text("Association Details", modifier = Modifier.testTag("association_details"))
+}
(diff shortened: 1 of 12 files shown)


app/src/main/java/com/github/swent/echo/compose/association/AssociationSearchBottomSheet.kt
```diff
@@ -0,0 +1,13 @@
...
+import androidx.compose.runtime.Composable
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun AssociationSearchBottomSheet(
+    onFullyExtended: () -> Unit,
+    onDismiss: () -> Unit,
+    searchEntryCallback: (String) -> Unit,
+    searched: String
+) {}
```
Greenade, line 13: another placeholder, seems like a lot of work to me for other branches, maybe you could request help for all the work.

Greenade (changes requested):
Thanks for your work ! Seems to me like there are a lot of placeholders, maybe you could ask for help on those as it looks like there's a lot of work to do left. Other than that everything looks great 💯 Just change the name of the variable I mentioned above and we're good 😄

app/src/main/java/com/github/swent/echo/compose/association/AssociationDetails.kt
```diff
@@ -0,0 +1,18 @@
...
+import com.github.swent.echo.data.model.Event
+
+@Composable
+fun AssociationDetails(
+    follow: (Association) -> Unit,
+    association: Association,
+    isFollowed: Boolean,
+    events: List<Event>
+) {
+    Text("Association Details", modifier = Modifier.testTag("association_details"))
```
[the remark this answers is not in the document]
ntissieres: the next PR arrive juste after this merge :)

app/src/main/java/com/github/swent/echo/compose/association/AssociationSearchBottomSheet.kt
ntissieres: I already finished everything on an intermediate PR ! Just pushing one by one and adding the tests ;)

Greenade (dismissed):

Greenade (approved):
Everything is good now 👍

PR #251, opened 2024-05-08
Create the logic for the associations in the app

ntissieres:
Read the .md file and don't hesitate to suggest modifications !

PR #294, opened 2024-05-15
Feature/association main screen

ntissieres:
The first thing a user sees is the active events of each association she/he follows. Reuse the ListDrawer. On top of this page, we have two buttons (or better MD component) : one with "Followed Associations" which leads to a list of followed associations by a user and one with "My Associations" which leads to a list of associations to which the user belongs. At the bottom right, there is a search button (like the one in HomeScreen) which allows the user to search an association.


app/src/main/java/com/github/swent/echo/compose/association/AssociationSubscriptionsScreen.kt
```diff
@@ -1,50 +0,0 @@
...
-import androidx.compose.runtime.getValue
-import androidx.compose.runtime.setValue
-import androidx.compose.ui.res.stringResource
-import com.github.swent.echo.R
-import com.github.swent.echo.ui.navigation.NavigationActions
-import com.github.swent.echo.viewmodels.user.UserProfileViewModel
-
-// This screen is a list of the subscriptions of a user to associations
-@Composable
-fun AssociationSubscriptionsScreen(
```
unglazedstamp, line 14: I believe this screen should be kept because it is used to unfollow associations easily and have an overview of the followed associations of the user.

app/src/main/java/com/github/swent/echo/compose/association/AssociationMainScreen.kt
```diff
@@ -18,15 +37,78 @@ fun AssociationMainScreen(
     addAssociationToFilter: (Association) -> Unit,
     followedAssociations: List<Association>,
     committeeAssociations: List<Association>,
+    eventsFilter: List<Association>,
+    isOnline: Boolean
+) {
+    Column(modifier = Modifier.fillMaxSize().testTag("association_main_screen")) {
```
unglazedstamp, line 43: This column should be scrollable so that we can see the list of event even if the associations list is expanded.

unglazedstamp (changes requested):
The code is well written. However I found some problems in the UI.
1. The description of the events in forced to be on the full width, this add some spacing between the words in some case and make the description difficult to read: see the general sport event description as example.
2. The join event button doesn't seem to be aligned to the right of the box.
3. I only see the first two letters of the associations in the associations menus.
4. The button to have the details of an association isn't easy to find for the user (maybe related to the two letters problem)
5. The title of the screen should be more precise or an explanation could be provided on the different parts of the screen (e.g. filter by association)

In general I don't really see why we need to filter the events by association on this screen since there is already another screen to filter events, the name of the button in the hamburger menu seems confusing. I think that the screen to easily unfollow associaitons is still necessary because it's more straightforward than navigating to each association detail page just to unfollow it.

ntissieres:
There are a lot of comments about the UI, which is a bit disappointing because I specifically made a mockup last Friday. I asked for it to be reviewed, but got no response, so I started the implementation on Sunday.
Anyway:
1. The problem isn't with this PR, but it's a good point you make, I'll have to adapt it in ListDrawer.
2. Once again, the problem doesn't come from this PR
3. That's right, I thought the association names were all small enough to fit in the box. I'm going to change that and let the width of the box adapt.
4. I thought about doing a special design to make it clear that it's clickable but with the amount of stuff I had to code, I just don't have the time to do it. But I can put the text in a button so that people understand even if it's ugly
5. Similarly, a better design would make it easier to understand. For now, I can add a title.
As for the page for filtering events by association, it's not possible to do that in the HomeScreen. This section of the application is there to highlight associations and their events, which is why I think it's still necessary. As for follow/unfollow on the list, this has already been discussed with Jonas here : #251.
I'm sorry for the slightly bitter response, but a lot of these comments could have been avoided if someone had told me before I spent 20 hours coding everything to do with associations. But that's OK, thanks anyway for your comments, which are still relevant.

app/src/main/java/com/github/swent/echo/compose/association/AssociationSubscriptionsScreen.kt
ntissieres: The overview is available on the AssociationMainScreen and massive unfollow is a pretty rare thing. It also avoid unintentional unfollow and drastically simplify the code + user flow (so we can use ListAssociation without trouble in other pages, which is used in the future also in AssociationSearch, without doing weird things with the UI to not have follow button on comitteeList or when we display all associations)

app/src/main/java/com/github/swent/echo/compose/association/AssociationMainScreen.kt
ntissieres: I can change that, thanks

app/src/main/java/com/github/swent/echo/compose/association/AssociationSubscriptionsScreen.kt
unglazedstamp: I though that last week we decided to keep the mass unfollow screen since I added a snackbar to undo a miss click, but I must have misunderstood the final decision.

unglazedstamp:
Sorry to come with problems from other parts of the code, I haven't found them before because I didn't review the screen with the list of events earlier. I will open a separate issue to address them.
1.2. see issue TBD
4. A quick solution could be to use TextButton (https://developer.android.com/develop/ui/compose/components/button#text_button), it would just change the style and color of the text so that it's clear that it's not just text.
5. I believe the design isn't bad but I found the content a bit surprising given the screen name, maybe we can discuss this today at the meeting to find a better name.

Now that I understand more what's the purpose of the screen, I see why you removed the old ones with the associations lists. I apologize for not giving an earlier feedback for this screen on Figma. I think your design is ok, what triggered my attention was that I though some features were removed from the app.

app/src/test/java/com/github/swent/echo/compose/association/AssociationMainScreenTest.kt
```diff
@@ -0,0 +1,128 @@
...
+                testAssociations[2],
+                "title 3",
+                "",
+                Location.EMPTY,
+                ZonedDateTime.now(),
+                ZonedDateTime.now(),
+                emptySet(),
+                0,
+                0,
+                0
```
alejandrocalles, line 76: Using `Event.EMPTY.copy(organizer = ..., title = ..., ...)` instead of instantiating an `Event` from scratch will provide better maintainability to this part of the code (e.g. if there are changes to the `Event` data class).

alejandrocalles (changes requested):
Thanks for working on this! It's truly a massive addition to the app. I requested a couple of comments, but they're mostly nitpicking.

app/src/main/java/com/github/swent/echo/compose/association/AssociationListScreen.kt
```diff
@@ -4,86 +4,36 @@ import androidx.compose.foundation.clickable
...
-            }
+    LazyColumn(modifier = Modifier.testTag("association_list_screen")) {
+        items(associationList.size) { index ->
+            AssociationListElement(
+                association = associationList[index],
+                onRowClicked = { onRowClicked(it) },
+                onAssociationClicked = { onAssociationClicked(it) },
+                eventsFilter
+            )
         }
```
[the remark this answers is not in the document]
unglazedstamp: would be nice but items is required here because AssociationListElement is a composable, so it cannt be put in a foreach.

app/src/test/java/com/github/swent/echo/compose/association/AssociationMainScreenTest.kt
unglazedstamp: That's true, I have done the change.

unglazedstamp (approved):

app/src/main/java/com/github/swent/echo/compose/association/AssociationListScreen.kt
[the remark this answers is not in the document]
alejandrocalles: You're right that it couldn't be directly put inside a `forEach`, but that's why it should be put within the scope of an `item { ... }`.
I made the changes locally and it worked for me, but this is not very relevant, so I'll approve it anyways.

app/src/test/java/com/github/swent/echo/compose/association/AssociationMainScreenTest.kt
alejandrocalles: Thanks!

alejandrocalles (approved):

PR #295, opened 2024-05-15
Feature/association search

ntissieres:
When the search button is clicked, it opens a bottom sheet with exactly the same functionality as Discover, but without the filters (which doesn't make sense for searching associations). A list with all associations is displayed in the background and updated according to the bottom sheet. If an item of the list is clicked, it leads the user to the specific AssociationPage.

### DIFF
--- a/app/src/main/java/com/github/swent/echo/compose/association/AssociationSearch.kt
+++ b/app/src/main/java/com/github/swent/echo/compose/association/AssociationSearch.kt
@@ -1,15 +1,23 @@
 package com.github.swent.echo.compose.association
 
-import androidx.compose.material3.ExperimentalMaterial3Api
-import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
-import androidx.compose.ui.Modifier
-import androidx.compose.ui.platform.testTag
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/github/swent/echo/compose/association/AssociationSearch.kt
```diff
@@ -1,15 +1,23 @@
...
+        onAssociationClicked = {
+            val nextPage = AssociationPage.DETAILS
+            nextPage.association = it
+            goTo(nextPage)
+        },
+        onRowClicked = {
+            val nextPage = AssociationPage.DETAILS
+            nextPage.association = it
+            goTo(nextPage)
+        },
```
unglazedstamp, line 20: Idea: these two callbacks seems similar, they could be abstracted in a function

unglazedstamp (changes requested):
Nice feature, thank you for working on this. Only one little thing that could be changed but not necessary.

unglazedstamp (dismissed):

unglazedstamp (approved):
Nice changes, thank you for working on this.

PR #296, opened 2024-05-15
Fix/profile composable

srsingh04:
All the bugs on the frontend for profile creation fixed.


octogradiste (changes requested):
Thanks so much for working on this! Good job on fixing the bugs! One thing to change is that the go back icon button. When createing a profile (user profile is null) we should either hide this button or log out the user and navigate back to the login screen.

octogradiste (approved):

PR #297, opened 2024-05-16
Feature/association details

ntissieres:
On top, we have the choice of two pages : a "Description" default page and a "Events" page.

Description
This page consists of the title and description of the association. On top right, there is a button to follow/unfollow the association.

Events
This page shows all events of the association. Reuse the ListDrawer already implemented in HomeScreen, filtered to display only events of the association.

### DIFF
--- a/app/src/main/java/com/github/swent/echo/compose/association/AssociationDetails.kt
+++ b/app/src/main/java/com/github/swent/echo/compose/association/AssociationDetails.kt
@@ -1,18 +1,156 @@
 package com.github.swent.echo.compose.association
 
+import androidx.compose.foundation.background
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.width
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.filled.Add
+import androidx.compose.material.icons.filled.Clear
+import androidx.compose.material3.Button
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/github/swent/echo/compose/association/AssociationDetails.kt
```diff
@@ -1,18 +1,169 @@
...
+                )
+                Text(association.url, modifier = Modifier.padding(5.dp))
+                 */
+            }
+            AssociationDetailsTab.EVENTS -> {
+                Spacer(modifier = Modifier.height(verticalSpace))
+                ListDrawer(events, "", "", isOnline)
+            }
+        }
+    }
```
alejandrocalles, line 168: The readability of this section could be improved. I believe there's two possible ways to do this:
1. Make a big switch on `associationDetailsTab`. This would have the benefit of being easy to read, but would lead to a lot of duplicated code, given the similarities between the two modes.
2. Given that the two modes share the same structure, make several small switches at the beginning to set the values that differ. For example,
```
val alignment = when (associationDetailsTab) {
    DESCRIPTION -> Alignment.CenterStart
    EVENTS -> Alignment.CenterEnd
}
```

alejandrocalles (changes requested):
Thanks again for working on the `Associations`! The code looks clean, I've only requested a small improvement to its readability.

app/src/main/java/com/github/swent/echo/compose/association/AssociationDetails.kt
ntissieres: That's a very good point ! I struggled to find a solution to this problem, that's some great idea you gived to me ! I'll create an issue for the next sprint in order to apply those changes ;)

alejandrocalles:
LGTM! Thanks :)

alejandrocalles (approved):

app/src/main/java/com/github/swent/echo/compose/association/AssociationDetails.kt
ntissieres: Here is the future issue : https://github.com/orgs/echo-swent/projects/1/views/3?pane=issue&itemId=63254196

PR #299, opened 2024-05-16
offline mode

srsingh04:
Offline mode clean PR as requested.

### DIFF
--- a/app/src/main/java/com/github/swent/echo/compose/components/OfflineUI.kt
+++ b/app/src/main/java/com/github/swent/echo/compose/components/OfflineUI.kt
@@ -4,14 +4,11 @@ import androidx.compose.animation.AnimatedVisibility
 import androidx.compose.animation.animateColorAsState
 import androidx.compose.animation.expandHorizontally
 import androidx.compose.animation.shrinkHorizontally
-import androidx.compose.foundation.layout.Arrangement
-import androidx.compose.foundation.layout.PaddingValues
-import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
-import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.material3.CenterAlignedTopAppBar
 import androidx.compose.material3.ExperimentalMaterial3Api
-import androidx.compose.material3.Scaffold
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/github/swent/echo/compose/components/OfflineUI.kt
```diff
@@ -0,0 +1,95 @@
...
+fun ConnectivityStatusBox(isConnected: Boolean, modifier: Modifier = Modifier) {
+    val backgroundColor by animateColorAsState(if (!isConnected) Color.LightGray else Color.Green)
+
+    val message = if (isConnected) R.string.Online_mode else R.string.Offline_mode
+    SmallTopAppBarExample(text = stringResource(id = message), color = backgroundColor)
+}
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun SmallTopAppBarExample(text: String, color: Color) {
```
alejandrocalles, line 71: This should probably have a more appropriate name, given that it's an actual composable used in the app, and not just an example.

app/src/main/java/com/github/swent/echo/compose/components/OfflineUI.kt
```diff
@@ -0,0 +1,95 @@
...
+    Scaffold(
```
alejandrocalles, line 72: Does it make sense to put a Scaffold here? If we want this to be only at the top of the app, maybe another composable makes more sense, such as a `Box`. That way there would be no need to mess with `ScrollContent`.

app/src/main/java/com/github/swent/echo/compose/components/OfflineUI.kt
```diff
@@ -0,0 +1,95 @@
...
+                title = { Text(text, fontSize = 15.sp) }
+            )
+        }
+    ) { innerPadding ->
+        ScrollContent(innerPadding)
+    }
+}
+
+@Composable
+fun ScrollContent(innerPadding: PaddingValues) {
```
alejandrocalles, line 89: This seems to be an empty composable. Does it really make sense to have it here?

alejandrocalles (changes requested):
Thanks for working on this! It'll substantially improve our UI. I've requested a couple of minor changes, but it should be good.

alejandrocalles (approved):

alejandrocalles:
Thanks, LGTM.

PR #302, opened 2024-05-16
Add and use custom serializer

octogradiste:


alejandrocalles (commented):
Thanks for working on this bug :) One small question: does it really makes sense to keep the `Semester` and `Section` fields nullable in the `UserProfile` data class given that there's a `UNKNOWN` object for each?

alejandrocalles:
It might be better to remove the `UNKNOWN` object altogether, as it's technically impossible for the Serializer to find invalid strings.

alejandrocalles:
LGTM. Thanks :)

alejandrocalles (approved):

PR #304, opened 2024-05-16
Feature/my events viewmodel

ntissieres:
Creating a view model in order to allow a user to join and leave an event

### DIFF
--- a/app/src/main/java/com/github/swent/echo/compose/components/EventInfoSheet.kt
+++ b/app/src/main/java/com/github/swent/echo/compose/components/EventInfoSheet.kt
@@ -1,6 +1,5 @@
 package com.github.swent.echo.compose.components
 
-import android.widget.Toast
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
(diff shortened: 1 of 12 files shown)


app/src/main/java/com/github/swent/echo/compose/components/EventInfoSheet.kt
```diff
@@ -148,24 +141,7 @@ fun EventInfoSheet(
...
-                            .show()
-                    },
-                    modifier = Modifier.testTag("join_button_event_info_sheet"),
-                ) {
-                    Text(
-                        text = stringResource(R.string.event_info_sheet_join_event_button_text),
-                        style = MaterialTheme.typography.labelLarge
-                    )
-                }
+                JoinEventButton(event, isOnline, 100.dp)
```
octogradiste, line 144: This should be something like 130.dp, otherwise the button text needs more than 1 line...

octogradiste (changes requested):
Excellent work! The code pleasant to read, the commits are clear, and the git history is easy to follow. Only one little visual change, the rest is all good!

octogradiste (approved):

PR #308, opened 2024-05-16
Integrate connectivity status

octogradiste:
Integrated the connectivity status + discovered that the loading animation was not properly centered on my phone, so I fixed that too.


srsingh04 (approved):
Ready to be merged! Thank you for making the respective changes and integrating it in the app.

PR #309, opened 2024-05-17
Prevent the Map from requesting location more than once.

### LINKED ISSUE #316: Location permissions should only be requested once per session.
Currently, as long as location permissions are not granted, they are requested each time the `MapDrawer` is recomposed. This should be fixed.

alejandrocalles:
Closes #316.

### DIFF
--- a/app/src/main/java/com/github/swent/echo/compose/components/HomeScreen.kt
+++ b/app/src/main/java/com/github/swent/echo/compose/components/HomeScreen.kt
@@ -32,7 +32,11 @@ import kotlinx.serialization.json.Json
  * view, and the map.
  */
 @Composable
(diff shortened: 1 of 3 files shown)


alejandrocalles:
Moving the handling of location permissions to the `AppNavigationHost` also sets the ground for #318.

violoncelloCH (approved):
thank you very much for working on this important usability improvement!

Just a small general remark on coding style: to make code as easy to read and understand as possible, it generally helps to avoid single-letter variable names, but rather use something expressive. It really helps if the chosen names for variables, functions and classes clearly express what they do/are.
But well done :)

alejandrocalles:
That's fair, I'll take care of this during the next sprint, as my tasks are closely related to the permissions. Thanks for your review :)

PR #313, opened 2024-05-17
Callback to retrigger the layout after joining/leaving an event

ntissieres:
Trying to solve the bug where the layout doesn't update the number of participants when we join/leave an event

### DIFF
--- a/app/src/test/java/com/github/swent/echo/viewmodels/HomeScreenViewModelTest.kt
+++ b/app/src/test/java/com/github/swent/echo/viewmodels/HomeScreenViewModelTest.kt
@@ -12,6 +12,7 @@ import com.github.swent.echo.data.model.UserProfile
 import com.github.swent.echo.data.repository.Repository
 import com.github.swent.echo.fakes.FakeAuthenticationService
 import io.mockk.coEvery
+import io.mockk.coVerify
 import io.mockk.every
 import io.mockk.mockk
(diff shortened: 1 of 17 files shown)


app/src/test/java/com/github/swent/echo/viewmodels/HomeScreenViewModelTest.kt
```diff
@@ -203,4 +205,37 @@ class HomeScreenViewModelTest {
...
+                    title = "New Event",
+                    description = "",
+                    location = Location("Location 2", MAP_CENTER.toLatLng()),
+                    startDate = ZonedDateTime.now(),
+                    endDate = ZonedDateTime.now(),
+                    tags = setOf(Tag("1", "new")),
+                    participantCount = 5,
+                    maxParticipants = 8,
+                    imageId = 0
+                )
```
alejandrocalles, line 228: I will do this next sprint, so it doesn't really matter that much, but it's preferable to use  `Event.EMTPY.copy(...)` instead of `Event(...)`. This goes for most other data classes as well.

That way there's no need to specify the fields that don't matter for the test, and it also improves the overall maintainability of the code (e.g. if changes are made to the `Event` data class there would be no need to refactor every single use of the `Event` constructor).

alejandrocalles (approved):
Thanks for working on this! I left a small, non-blocking comment, but other than that everything looks perfect.

PR #317, opened 2024-05-17
🐛 HIGH VALUE BUGFIX AFTER 3 HOURS OF DEBUGGING: properly refetch events *after* joining/leaving event 

violoncelloCH:
After spending 3 hours debugging why #313 didn't have it's intended effect when taken together with #310, I finally figured out how to make this work thanks to the help of @unglazedstamp 

The refresh action being launched in it's own coroutine, it refreshed the viewmodel before the coroutine which made the call to the repository for joining/leaving (which updates the even in the cache) finished. Therefore the applied changes were not visible. This change makes sure the refresh of the viewmodel and consequently the UI state is only done after the call to the repository finished.

~:warning: requires #310 to be merged first -> only look at the last commit for now.~ EDIT: merged


alejandrocalles (approved):
Thanks for fixing this! It all seems to work now, and the code is clean.

PR #320, opened 2024-05-17
Fix/profile creation

srsingh04:
Bug fix for tags


octogradiste (approved):
Indead fixes the bug! Thanks a lot!

PR #323, opened 2024-05-19
subtags in event filtering (homescreen)

Greenade:
search mode and follow mode now filter events in the subtags of selected tags as well

### DIFF
--- a/app/src/main/java/com/github/swent/echo/viewmodels/HomeScreenViewModel.kt
+++ b/app/src/main/java/com/github/swent/echo/viewmodels/HomeScreenViewModel.kt
@@ -97,6 +97,7 @@ constructor(
     val followedTags = _followedTags.asStateFlow()
     private val _selectedTagId = MutableStateFlow<String?>(null)
     val selectedTagId = _selectedTagId.asStateFlow()
+    private var followedTagFilter = listOf<Tag>()
     private var _searchMode = MutableStateFlow(false)
     val searchMode = _searchMode.asStateFlow()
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/github/swent/echo/viewmodels/HomeScreenViewModel.kt
```diff
@@ -203,11 +206,30 @@ constructor(
...
-        filterTagSet = allTagSet.filter { tag -> areWordsInTag(tag, filterWordList) }.toSet()
+        val temp = allTagSet.filter { tag -> areWordsInTag(tag, filterWordList) }
+        filterTagSet = getTagsAndSubTags(temp)
+
         updateSearchMode()
         filterEvents()
     }
 
+    private fun getTagsAndSubTags(initialTagList: List<Tag>): Set<Tag> {
+        if (initialTagList.isEmpty()) return setOf()
```
srsingh04, line 217: knitpick: intsead of setOf(), maybe emptySet() would be better?

srsingh04 (approved):
Thank you for the changes. Looks good to be merged.

app/src/main/java/com/github/swent/echo/viewmodels/HomeScreenViewModel.kt
Greenade: I never used emptySet() when initializing sets (or lists with emptyList()) so to make the code uniform I think I should keep it like that tbh. If you want we could still make every initialization emptySet/List later on though

PR #326, opened 2024-05-19
add section and semester tags button to the event creation/edition screen

unglazedstamp:


app/src/main/java/com/github/swent/echo/compose/event/EventTagEntry.kt
```diff
@@ -35,9 +42,28 @@ fun EventTagEntry(
     onTagSelected: (tag: Tag) -> Unit,
     onTagDeleted: (tag: Tag) -> Unit
 ) {
-    var dialogVisible by remember { mutableStateOf(false) }
     Column(modifier = Modifier.fillMaxWidth().padding(EVENT_PADDING_BETWEEN_INPUTS)) {
         EventEntryName(stringResource(R.string.edit_event_screen_tags))
+        Row {
```
ntissieres, line 47: Some repetitions here, you could create a list of Pair<String, String> and then iterates over it

ntissieres (changes requested):
Qualitative code, thanks for your work !  Just a little change to avoid repetition ;)

app/src/main/java/com/github/swent/echo/compose/event/EventTagEntry.kt
unglazedstamp: Thanks a lot for the suggestions, I changed that.

ntissieres (dismissed):
Thanks for the update ! Nice work ;)

ntissieres (approved):
Rebasing approval

PR #328, opened 2024-05-21
Help page

octogradiste:
A simple help page such that the user can quickly get started with our app!


app/src/main/res/values/strings.xml
```diff
@@ -105,5 +105,18 @@
     <string name="association_details_events">Events</string>
     <string name="list_drawer_leave_event">Leave Event</string>
     <string name="event_successfully_left">Event successfully left!</string>
+    <string name="help_screen_how_can_i_find_events">How can I find events?</string>
+    <string name="help_screen_how_can_i_find_events_description">Good question! By default you see all events that match your interests. You can filter by only one of your interests by tapping on the chips on the top of the map.\n\nTap on the search icon in the bottom left part of the screen to open the search options. This mode will consider all events, not only your interests. Perfect for discovering new events on the campus! In the top right corner you have an icon button to cancel the search.</string>
+    <string name="help_screen_create_event">I want to create an event</string>
+    <string name="help_screen_create_event_description">Great! You have to options: either you open the hamburger menu and tap on \'Create Event\' or you simply long press on the map. Then fill in the details and save it! It\'s this simple.</string>
```
Greenade, line 111: small typo : "You have two options" instead of "You have to options"
small detail : "It's that simple" instead of "it's this simple" is more common

app/src/main/res/values/strings.xml
```diff
@@ -105,5 +105,18 @@
...
+    <string name="help_screen_following_associations">Following associations</string>
+    <string name="help_screen_following_associations_description">Follow association to see what events they are hosting! To see all associations you are following open the hamburger menu and tap on \'Association\'.</string>
+    <string name="help_screen_my_interest_changed">My interest changed</string>
```
Greenade, line 114: interests plural would make more sense

app/src/main/java/com/github/swent/echo/compose/components/HamburgerMenuDrawerSheet.kt
```diff
@@ -95,17 +96,17 @@ fun HamburgerMenuDrawerSheet(
...
-
              */
             NavigationItem(
                 title = stringResource(R.string.hamburger_associations),
                 selectedIcon = Icons.Filled.Star,
                 navOnClick = { navActions.navigateTo(Routes.ASSOCIATIONS) }
             ),
+            NavigationItem(
+                title = stringResource(id = R.string.hamburger_help),
+                selectedIcon = Icons.Filled.Build,
```
Greenade, line 107: another icon would make more sense in my opinion, I suggest [the "help" icon from google](https://fonts.google.com/icons?icon.platform=android&selected=Material+Symbols+Outlined:help:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=help)

Greenade (changes requested):
Thank you for your work ! Very clean code ! I made suggestions about the icon, some typos and text we could change. I think we could also add a part about the list mode in my opinion, something like : "I can't see events well on the map" - "No worries ! You can change the display of events to a list by clicking the button in the upper right of the screen. You can even sort it by date or distance."

Greenade (approved):
Perfect work ! That you for addressing the changes I requested. 💯

PR #336, opened 2024-05-22
Feature/adding pager

ntissieres:
Creating a pager which allow the user to switch between two or more pages by either click on the title of the pages or by swiping on the screen. Including it inside AssociationDetails and MyEventsScreen

### DIFF
--- a/app/src/main/java/com/github/swent/echo/compose/components/Pager.kt
+++ b/app/src/main/java/com/github/swent/echo/compose/components/Pager.kt
@@ -1,0 +1,88 @@
+package com.github.swent.echo.compose.components
+
+import androidx.compose.foundation.ExperimentalFoundationApi
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.fillMaxHeight
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.wrapContentHeight
+import androidx.compose.foundation.pager.HorizontalPager
+import androidx.compose.foundation.pager.rememberPagerState
+import androidx.compose.material3.MaterialTheme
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/github/swent/echo/compose/components/Pager.kt
```diff
@@ -0,0 +1,88 @@
...
+            }
+        }
+        Spacer(modifier = Modifier.height(spaceBetweenTitleAndItem))
+        HorizontalPager(
+            state = pagerState,
+            modifier = Modifier.fillMaxHeight(),
+            verticalAlignment = Alignment.Top,
+        ) {
+            content[it].second()
+        }
```
alejandrocalles, line 86: This is simply really high quality code. I appreciate it.

app/src/main/java/com/github/swent/echo/compose/myevents/MyEventsScreen.kt
```diff
@@ -47,77 +28,31 @@ fun MyEventsScreen(myEventsViewModel: MyEventsViewModel, navActions: NavigationA
...
+        Box(modifier = Modifier.padding(it)) {
+            Pager(
+                listOf(
+                    Pair(stringResource(R.string.my_events_joined_events)) {
+                        ListDrawer(
+                            joinedEventsList,
+                            "",
+                            "",
+                            isOnline,
+                            myEventsViewModel::refreshEvents
```
alejandrocalles, line 43: [Nitpicking] Not sure if this is relevant to this PR, but it feels odd to pass empty string to the `ListDrawer` in order to (I assume) not show any section or semester. I believe it could be better (as in easier to understand for someone reading the code) to make both parameters in `ListDrawer` nullable, What do you think?
My assumptions might be wrong, so feel free to correct me.

alejandrocalles (changes requested):
Really clean PR, thank you so much for this! I left a small comment, but it's more of an inquiry than a request for changes.

app/src/main/java/com/github/swent/echo/compose/myevents/MyEventsScreen.kt
ntissieres: That's a good point ! Thanks for highlighting this. I will open an issue in order to change the way ListDrawer works, so don't have to pass those empty strings

alejandrocalles:
All good. Thanks!

alejandrocalles (approved):

PR #346, opened 2024-05-23
user profile picture in the backend

unglazedstamp:
add user profile picture support in the repository and supabase

### DIFF
--- a/app/src/main/java/com/github/swent/echo/data/repository/RepositoryImpl.kt
+++ b/app/src/main/java/com/github/swent/echo/data/repository/RepositoryImpl.kt
@@ -5,6 +5,7 @@ import com.github.swent.echo.data.model.Association
 import com.github.swent.echo.data.model.Event
 import com.github.swent.echo.data.model.Tag
 import com.github.swent.echo.data.model.UserProfile
+import com.github.swent.echo.data.repository.datasources.FileCache
 import com.github.swent.echo.data.repository.datasources.LocalDataSource
 import com.github.swent.echo.data.repository.datasources.RemoteDataSource
 import java.time.ZonedDateTime
@@ -12,7 +13,8 @@ import java.time.ZonedDateTime
 class RepositoryImpl(
     private val remoteDataSource: RemoteDataSource,
     private val localDataSource: LocalDataSource,
-    private val networkService: NetworkService
+    private val networkService: NetworkService,
+    private val fileCache: FileCache
 ) : Repository {
 
     private val isOffline: () -> Boolean = { !networkService.isOnlineNow() }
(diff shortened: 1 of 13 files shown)


app/src/main/java/com/github/swent/echo/data/repository/RepositoryImpl.kt
```diff
@@ -236,6 +237,27 @@ class RepositoryImpl(
         remoteDataSource.deleteUserProfile(userProfile)
         localDataSource.deleteUserProfile(userProfile.userId)
     }
+
+    override suspend fun getUserProfilePicture(userId: String): File? {
+        if (isOffline()) {
+            return null
+        }
+        return remoteDataSource.getUserProfilePicture(userId)
+    }
```
violoncelloCH, line 246: I think it would make sense to cache images locally on disk (to avoid network delay). Additionally, this would allow to serve images from cache even if offline.

The following documentation/tutorial could help implementing a LRU disk cache for images: https://developer.android.com/topic/performance/graphics/cache-bitmap

An even simpler way, as you already store the images to the cache directory, could be to simply check there if the picture already exists and when it has been saved. If it exists and that has not been too long ago, return the image from local storage directly, otherwise fetch from supabase. 

Also when offline, you could still see if there is a cached copy of the image available and return that.

You can take a look at how I did the caching/synchronisation for the database records above :)

app/src/main/java/com/github/swent/echo/data/supabase/SupabaseDataSource.kt
```diff
@@ -279,4 +283,27 @@ class SupabaseDataSource(private val supabase: SupabaseClient) : RemoteDataSourc
     override suspend fun deleteUserProfile(userProfile: UserProfile) {
         supabase.from("user_profiles").delete { filter { eq("user_id", userProfile.userId) } }
     }
+
+    override suspend fun getUserProfilePicture(userId: String): File? {
+        val outputFile =
+            File.createTempFile(userId, ".png", null) // the cache directory is used by default
+        supabase.storage
+            .from("user-profile-picture")
+            .downloadAuthenticatedTo("$userId.png", outputFile)
```
violoncelloCH, line 292: I fear that generally profile pictures are real images taken by a camera and not graphics with a relatively limited set of colours. Therefore jpg with it's compression would be preferable over png. What do you think; could this be changed?

violoncelloCH (commented):
Thanks for working on support for images! this will be a good base to have images for other usecases in the app as well later on :)
Well written!

I've just got two points to consider below.

app/src/main/java/com/github/swent/echo/data/repository/RepositoryImpl.kt
unglazedstamp: These ideas are good. I already tried to save the images on a specific cache sub-folder, but I had issues with the permissions and even the app cache path returned by `context.cacheDir` wasn't writable. The only way I could achieve to  save a file into the cache is by using the default cache directory of the method `File .createTempFile`. The problem with this solution is that I don't know where the default cache directory is, so I can't look into this directory before fetching the picture from Supabase.
I could address these issues in a future sprint, but given the time constraint I prefer to keep it like this even if it's not an optimal solution. Thanks for pointing to some helpful documentation to solve this problem.

unglazedstamp:
After some in person discussion, I will use the new hilt module for file caching (#357) to improve the way the repository getter and setter for the user profile picture work.

violoncelloCH (approved):
thanks for implementing the changes! LGTM now :)

PR #348, opened 2024-05-23
Imporve register UX

### LINKED ISSUE #101: Differentiate or merge the login and register screens.
Potential solutions:
- Merge both screens into a single one with both a login and a register button.
- Make it so that the text fields are preserved when switching from the register screen to the login screen.
- Create a different design for the register screen.

octogradiste:
- Added a confirm-password field to better distinguish between the login and register screen. Closes #101.
- Added validation and error texts for the register screen.
- Pressing enter moves to the next text field.

### DIFF
--- a/app/src/main/java/com/github/swent/echo/compose/authentication/AuthenticationForm.kt
+++ b/app/src/main/java/com/github/swent/echo/compose/authentication/AuthenticationForm.kt
@@ -25,6 +25,12 @@ import androidx.compose.ui.text.input.VisualTransformation
 import androidx.compose.ui.unit.dp
 import com.github.swent.echo.R
 
+enum class AuthenticationFormError(val id: Int) {
+    EMPTY_EMAIL(R.string.authentication_error_email_is_required),
+    PASSWORD_TOO_SHORT(R.string.authentication_password_too_short),
+    PASSWORD_DO_NOT_MATCH(R.string.authentication_error_passwords_do_not_match),
+}
+
 /**
  * A form for authenticating a user. It contains fields for the user's email and password.
  *
@@ -53,6 +59,13 @@ fun AuthenticationForm(
     var passwordError by remember { mutableStateOf("") }
     var confirmError by remember { mutableStateOf("") }
 
+    val minPasswordLength = 8
+    val spaceBetweenTextFields = 8.dp
+
+    val emailErrorText = stringResource(AuthenticationFormError.EMPTY_EMAIL.id)
+    val passwordTooShort = stringResource(AuthenticationFormError.PASSWORD_TOO_SHORT.id)
+    val passwordDoNotMatch = stringResource(AuthenticationFormError.PASSWORD_DO_NOT_MATCH.id)
+
     Column(
         modifier = Modifier.fillMaxWidth(),
         horizontalAlignment = Alignment.CenterHorizontally,
@@ -67,7 +80,7 @@ fun AuthenticationForm(
             },
             errorText = emailError,
         )
-        Spacer(modifier = Modifier.padding(8.dp))
+        Spacer(modifier = Modifier.padding(spaceBetweenTextFields))
         AuthenticationTextField(
             modifier = Modifier.fillMaxWidth().testTag("password-field"),
             label = stringResource(R.string.authentication_form_password_label),
@@ -80,7 +93,7 @@ fun AuthenticationForm(
             isPasswordField = true,
         )
         if (confirmPassword) {
-            Spacer(modifier = Modifier.padding(8.dp))
+            Spacer(modifier = Modifier.padding(spaceBetweenTextFields))
             AuthenticationTextField(
                 modifier = Modifier.fillMaxWidth().testTag("confirm-password-field"),
                 label = stringResource(R.string.authentication_form_confirm_password_label),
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/github/swent/echo/compose/authentication/AuthenticationForm.kt
```diff
@@ -28,47 +31,133 @@ import com.github.swent.echo.R
...
-            singleLine = true
+            onValueChange = {
+                emailError = ""
+                email = it
+            },
+            errorText = emailError,
         )
-        Spacer(modifier = Modifier.padding(16.dp))
-        OutlinedTextField(
+        Spacer(modifier = Modifier.padding(8.dp))
```
ntissieres, line 70: magic number, it's better to specify 8.dp in a val at the beginning of the composable

app/src/main/java/com/github/swent/echo/compose/authentication/AuthenticationForm.kt
```diff
@@ -28,47 +31,133 @@ import com.github.swent.echo.R
...
+
+                    if (password.length < 8) {
+                        passwordError = "Password must be at least 8 characters"
+                    }
+
+                    if (confirmPassword && password != confirm && passwordError.isEmpty()) {
+                        confirmError = "Passwords do not match"
+                    }
+
+                    if (emailError.isEmpty() && passwordError.isEmpty() && confirmError.isEmpty()) {
```
ntissieres, line 114: I feel like having 3 differents strings to handle the errors is a bit repetitive and non-practical if we have to add new errors or modifying the text message in the future. Also using strings to see errors is a bit weird. I think it would be better to create an enum with the different type of errors that contains a message, and then create a unique set of errors that you feed in the above if's. This way, you just have to test if this set is empty in the last condition, and the code is easier to maintain.

app/src/main/java/com/github/swent/echo/compose/authentication/AuthenticationForm.kt
```diff
@@ -28,47 +31,133 @@ import com.github.swent.echo.R
...
+                if (isPasswordField)
+                    KeyboardOptions(
+                        keyboardType = KeyboardType.Password,
+                        imeAction = ImeAction.Next
+                    )
+                else KeyboardOptions(keyboardType = KeyboardType.Text, imeAction = ImeAction.Next),
+            singleLine = true,
+        )
+        if (errorText.isNotEmpty()) {
+            Spacer(modifier = Modifier.padding(4.dp))
```
ntissieres, line 155: Same comment for Spacer

app/src/main/java/com/github/swent/echo/compose/authentication/RegisterScreen.kt
```diff
@@ -48,8 +51,10 @@ fun RegisterScreen(registerViewModel: RegisterViewModel, navActions: NavigationA
         }
     }
 
+    val scrollState = rememberScrollState()
+
     Column(
-        modifier = Modifier.testTag("register-screen"),
+        modifier = Modifier.testTag("register-screen").fillMaxHeight().verticalScroll(scrollState),
```
ntissieres, line 57: It is generally preferable to use a LazyColumn if you want a scrollable Column, but because you have only a few items, the difference in performance is low. It's not so important in this context but think about it the next time ;) 
source : https://stackoverflow.com/questions/73734918/android-compose-difference-between-lazycolumn-and-column-with-verticalscroll#:~:text=A%20LazyColumn%20is%20a%20vertically,the%20classic%20Android%20View%20system.&text=A%20Column%20will%20show%20each,a%20LinearLayout%20with%20vertical%20orientation.

app/src/main/java/com/github/swent/echo/compose/authentication/AuthenticationForm.kt
```diff
@@ -28,47 +31,133 @@ import com.github.swent.echo.R
...
+        }
+        Spacer(modifier = Modifier.padding(8.dp))
         ElevatedButton(
             enabled = isOnline,
             modifier = Modifier.fillMaxWidth().testTag("action-button"),
-            onClick = { onAuthenticate(email, password) },
+            onClick = {
+                if (validate) {
+                    if (email.isEmpty()) {
+                        emailError = "Email is required"
```
ntissieres, line 103: This error message could be a stringResource

app/src/main/java/com/github/swent/echo/compose/authentication/AuthenticationForm.kt
```diff
@@ -28,47 +31,133 @@ import com.github.swent.echo.R
...
+                    }
+
+                    if (password.length < 8) {
+                        passwordError = "Password must be at least 8 characters"
```
ntissieres, line 107: same here

app/src/main/java/com/github/swent/echo/compose/authentication/AuthenticationForm.kt
```diff
@@ -28,47 +31,133 @@ import com.github.swent.echo.R
...
+                    }
+
+                    if (confirmPassword && password != confirm && passwordError.isEmpty()) {
+                        confirmError = "Passwords do not match"
```
ntissieres, line 111: same here

ntissieres (changes requested):
Good code overall but needs some little changes to avoid magic numbers and add string resources !

ntissieres (dismissed):

ntissieres (approved):
Nice ! Thanks for the corrections ;)

PR #349, opened 2024-05-23
Update/association UI update

ntissieres:


app/src/main/java/com/github/swent/echo/compose/association/AssociationListScreen.kt
```diff
@@ -23,16 +18,12 @@ import com.github.swent.echo.data.model.Association
...
-    eventsFilter: List<Association>
 ) {
     LazyColumn(modifier = Modifier.testTag("association_list_screen")) {
         items(associationList.size) { index ->
             AssociationListElement(
                 association = associationList[index],
-                onRowClicked = { onRowClicked(it) },
-                onAssociationClicked = { onAssociationClicked(it) },
-                eventsFilter
+                onRowClicked = { onRowClicked(it) }
```
octogradiste, line 26: You could simply use `onRowClicked = onRowClicked`

app/src/main/java/com/github/swent/echo/compose/association/AssociationListScreen.kt
```diff
@@ -43,39 +34,23 @@ fun AssociationListScreen(
...
-                        .testTag("selected_icon_${association.name}")
-                )
-            }
+            Text(
+                text = association.name,
+                textAlign = TextAlign.Center,
+                modifier =
+                    Modifier.padding(boxInsidePadding)
+                        .testTag("association_name_button_${association.name}")
+            )
```
octogradiste, line 53: You could move this inside `headlineContent` avoids having `headlineContent = {}`...

app/src/main/java/com/github/swent/echo/compose/components/Pager.kt
```diff
@@ -33,8 +35,8 @@ import kotlinx.coroutines.launch
  */
 @OptIn(ExperimentalFoundationApi::class)
 @Composable
-fun Pager(content: List<Pair<String, @Composable () -> Unit>>) {
-    val pagerState = rememberPagerState(pageCount = { content.size })
+fun Pager(content: List<Pair<String, @Composable () -> Unit>>, initialPage: Int? = 0) {
```
octogradiste, line 38: You could use `initialPage: Int = 0` as you never need initial page to be null.

app/src/main/java/com/github/swent/echo/viewmodels/association/AssociationViewModel.kt
```diff
@@ -109,41 +73,26 @@ constructor(
...
-            allEvents.filter {
-                _followedAssociations.value.contains(it.organizer) ||
-                    _committeeAssociations.value.contains(it.organizer)
-            }
-        } else {
-            allEvents.filter { _eventsFilter.value.contains(it.organizer) }
-        }
+    fun setCurrentAssociationPage(
+        association: Association,
+        initialPage: Int? = _initialPage.value
```
octogradiste, line 89: Here `initialPage` could also be of type `Int` and not `Int?`.

octogradiste (approved):
Thanks so much for the association UI improvement! The code is very clean and easy to read. I left some comments on stuff you could improve, but as it's rather minor I already approved. Up to you to make the changes or merge directly :)

app/src/main/java/com/github/swent/echo/compose/association/AssociationListScreen.kt
ntissieres: Right ! it's more logical

app/src/main/java/com/github/swent/echo/compose/components/Pager.kt
ntissieres: I thought the ? was the way to allow a function to not take argument, I didn't know that we could do it this way ! Thanks for highlighting this ;)

octogradiste (approved):
LGTM!!

PR #352, opened 2024-05-24
Feature/many selected tags

Greenade:
made it possible to select more than one followed tags in the homescreen

### DIFF
--- a/app/src/main/java/com/github/swent/echo/viewmodels/HomeScreenViewModel.kt
+++ b/app/src/main/java/com/github/swent/echo/viewmodels/HomeScreenViewModel.kt
@@ -95,8 +95,8 @@ constructor(
     private lateinit var semesterTags: List<Tag>
     private val _followedTags = MutableStateFlow<List<Tag>>(listOf())
     val followedTags = _followedTags.asStateFlow()
-    private val _selectedTagId = MutableStateFlow<String?>(null)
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/github/swent/echo/viewmodels/HomeScreenViewModel.kt
```diff
@@ -125,13 +125,19 @@ constructor(
...
         if (_followedTags.value.contains(tag)) {
-            if (_selectedTagId.value == tag.tagId) {
-                _selectedTagId.value = null
-                followedTagFilter = getTagsAndSubTags(_followedTags.value).toList()
+            if (_selectedTagIds.value.contains(tag.tagId)) {
+                _selectedTagIds.value = _selectedTagIds.value.minus(tag.tagId)
             } else {
-                _selectedTagId.value = tag.tagId
-                followedTagFilter = getTagsAndSubTags(listOf(tag)).toList()
+                _selectedTagIds.value = _selectedTagIds.value.plus(tag.tagId)
```
violoncelloCH, line 131: hint: you could use `-` and `+` here instead of the explicit `.minus()` and `.plus()` methods :)

violoncelloCH (approved):
LGTM! Thanks for working on this: it makes the quick tag filtering way more useful :)

PR #353, opened 2024-05-24
Show user's distance to events in the HomeScreen.

### LINKED ISSUE #318: Implement a GPS service to get the user position
Takes a list of events and returns it sorted by closest

alejandrocalles:
Should close #318.

If permissions are enabled, every `EventListitem` in the `HomeScreen` should show how far the user is from the event.

Distance should also be shown in the EventInfoSheet.

Should allow users to sort events by distance in the `HomeScreen`.


alejandrocalles:
Closing this because the UI has changed too much for this to be relevant. A new branch will be created from #373 instead.

PR #354, opened 2024-05-24
Improve event list item

octogradiste:
This simplifies and cleans up the UI for the event list item.

It also adds a `viewOnMap` and a  `modifyEvent` callback. Those will need to be configured in the `HomeScreen` and will require some changes in the view model. This is for another PR.


app/src/main/java/com/github/swent/echo/compose/components/ListDrawer.kt
```diff
@@ -25,261 +30,225 @@ import androidx.compose.runtime.mutableStateOf
...
-                                    }
-                                )
-                                    colorClass
-                                else color2ndFloor
-                            )
+                Column(
+                    verticalArrangement = Arrangement.Center,
+                    horizontalAlignment = Alignment.CenterHorizontally,
                 ) {
+                    Icon(painterResource(R.drawable.full), contentDescription = "Group")
```
unglazedstamp, line 167: This icon description could be more explicit so that the number at its right makes more sense when screen readers read it, e.g. "number of participants".

unglazedstamp (commented):
Nice code, really pleasant to read. The comments are clear and helpful. I launched the app and when I go to the created events screen I can't edit the events created by the test user (e.g. we fish).

unglazedstamp (changes requested):
See above.

octogradiste:
> Is it intended ? Also some callbacks and parameters aren't used yet but I suppose that's because other team members are working on these features.

Yeah, there is @alejandrocalles working on #353 which modifies the home screen and home screen view model. So, connecting the new callbacks with the home screen view model would create git conflicts. We should connect the callbacks only once he is finished with his PR.

unglazedstamp (approved):

unglazedstamp:
Thanks for the improvements !

PR #355, opened 2024-05-26
add url and relatedTag to Associations and make Event and UserProfile only include a header of Association(s)

### LINKED ISSUE #264: Adapt the dataclass Association
We already have a dataclass called Association. Currently, it has the following attributes : an "associationId", a "name" and a "description". We can add to this :
- "relatedTags: Set<Tag>", set of all tags related to the association
- "associationURL : URL", the most important URL of the association (website, facebook, instagram, ...)

violoncelloCH:
closes #264

This extends the Association dataclass to hold two additional attributes: `url` and `relatedTags`. While the url is a simple string field in the db, the relatedTags are stored as a relation between the Association and the Tag. 
Since Associations were embedded inside Events as well as UserProfile data classes, this would have meant fetching all the Tags related to the Associations whenever we want a UserProfile (or an Event). To be precise, we talk about multiple four-way nested joins across a total of 8 tables to fetch a UserProfile. In order to simplify this, together with @octogradiste we decided to not embed Associations in Events and UserProfiles anymore. Instead we just embed a simplified object, called `AssociationHeader`, only containing the id and the name of the Association, analogous to how we already do it for the `EventCreator` on Events.

### DIFF
--- a/app/src/main/java/com/github/swent/echo/data/room/RoomLocalDataSource.kt
+++ b/app/src/main/java/com/github/swent/echo/data/room/RoomLocalDataSource.kt
@@ -6,6 +6,7 @@ import com.github.swent.echo.data.model.Tag
 import com.github.swent.echo.data.model.UserProfile
 import com.github.swent.echo.data.repository.datasources.LocalDataSource
 import com.github.swent.echo.data.room.entity.AssociationRoom
+import com.github.swent.echo.data.room.entity.AssociationTagCrossRef
 import com.github.swent.echo.data.room.entity.EventRoom
 import com.github.swent.echo.data.room.entity.EventTagCrossRef
 import com.github.swent.echo.data.room.entity.JoinedEventRoom
@@ -14,7 +15,7 @@ import com.github.swent.echo.data.room.entity.UserProfileAssociationSubscription
 import com.github.swent.echo.data.room.entity.UserProfileCommitteeMemberCrossRef
 import com.github.swent.echo.data.room.entity.UserProfileRoom
 import com.github.swent.echo.data.room.entity.UserProfileTagCrossRef
-import com.github.swent.echo.data.room.entity.toAssociationRoomList
+import com.github.swent.echo.data.room.entity.toAssociations
 import com.github.swent.echo.data.room.entity.toTagList
 import com.github.swent.echo.data.room.entity.toTagRoomList
 import java.time.ZonedDateTime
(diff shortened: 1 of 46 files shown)


app/src/main/java/com/github/swent/echo/data/room/RoomLocalDataSource.kt
```diff
@@ -97,7 +131,6 @@ class RoomLocalDataSource @Inject constructor(db: AppDatabase) : LocalDataSource
             events.flatMap { event -> event.tags.map { EventTagCrossRef(event.eventId, it.tagId) } }
 
         eventDao.deleteAllEventTagCrossRefsForEvents(events.map { it.eventId })
-        associationDao.insertAll(associations.map { AssociationRoom(it) })
```
octogradiste, line 100: The variable `associations` is now unused and could be deleted.

app/src/main/java/com/github/swent/echo/data/room/RoomLocalDataSource.kt
```diff
@@ -42,20 +43,54 @@ class RoomLocalDataSource @Inject constructor(db: AppDatabase) : LocalDataSource
...
     override suspend fun setAssociations(associations: List<Association>) {
+        val tags = associations.flatMap { it.relatedTags }.toSet()
+        val crossRefs =
+            associations.flatMap { association ->
+                association.relatedTags.map {
+                    AssociationTagCrossRef(association.associationId, it.tagId)
+                }
+            }
+
+        eventDao.deleteAllEventTagCrossRefsForEvents(associations.map { it.associationId })
```
octogradiste, line 90: I think this should be `deleteAllAssociationTagCrossRefsForAssociation`.

octogradiste (changes requested):
Massive work! Thank you so much! The code is very clean and easy to understand. Two things to change. One of them, I'm not so sure, but looks like a bug to me...

app/src/main/java/com/github/swent/echo/data/room/RoomLocalDataSource.kt
violoncelloCH: nice catch! thanks!
you're absolutely right! I made a copy paste error here :see_no_evil:

app/src/main/java/com/github/swent/echo/data/room/RoomLocalDataSource.kt
violoncelloCH: you're right, I'll remove that :)

octogradiste (approved):
LGTM!

PR #357, opened 2024-05-27
Image cache

octogradiste:
Interface to save and get bytes from the cache. Will be used to save the profile image to the cache. Note that you can inject it in any part of the code using Hilt!

### DIFF
--- a/app/src/main/java/com/github/swent/echo/data/repository/datasources/FileCache.kt
+++ b/app/src/main/java/com/github/swent/echo/data/repository/datasources/FileCache.kt
@@ -22,6 +22,7 @@ interface FileCache {
      * Delete the file by name.
      *
      * @param name The name of the file to delete.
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/github/swent/echo/data/repository/datasources/FileCache.kt
```diff
@@ -0,0 +1,27 @@
+package com.github.swent.echo.data.repository.datasources
+
+/** Interface for file caching. */
+interface FileCache {
+
+    /**
+     * Get the file content by name.
+     *
+     * @param name The name of the file to get.
```
unglazedstamp, line 9: It could be useful to add a `@return` section in this comment to specify when this function shoud return `null`

unglazedstamp (approved):
Very nice code ! Thanks for this hilt module which will allow easy caching of the files. I found one little optional comment improvement.

PR #358, opened 2024-05-27
user profile picture in the create profile screen and the hamburger menu

unglazedstamp:
need #346 merged first

### DIFF
--- a/app/src/main/java/com/github/swent/echo/compose/authentication/CreateProfile.kt
+++ b/app/src/main/java/com/github/swent/echo/compose/authentication/CreateProfile.kt
@@ -569,7 +569,8 @@ fun PictureTransformer(
                         translationX = offset.x,
                         translationY = offset.y
                     )
-                    .transformable(state = state),
+                    .transformable(state = state)
+                    .testTag("profile-picture-image"),
             painter = BitmapPainter(picture.asImageBitmap()),
             contentDescription = ""
         )
--- a/app/src/main/java/com/github/swent/echo/compose/authentication/CreateProfile.kt
+++ b/app/src/main/java/com/github/swent/echo/compose/authentication/CreateProfile.kt
@@ -199,6 +199,9 @@ fun ProfileCreationUI(
 ) {
     val scope = rememberCoroutineScope()
     val snackbarHostState = remember { SnackbarHostState() }
+    val spacerHeight = 5.dp
+    val spaceBetweenTags = 5.dp
+    val contentPadding = 16.dp
 
     Scaffold(
         topBar = {
@@ -226,7 +229,10 @@ fun ProfileCreationUI(
         ) {
             Column(
                 modifier =
-                    modifier.fillMaxSize().padding(16.dp).verticalScroll(rememberScrollState())
+                    modifier
+                        .fillMaxSize()
+                        .padding(contentPadding)
+                        .verticalScroll(rememberScrollState())
             ) {
                 Row {
                     Column {
@@ -247,7 +253,7 @@ fun ProfileCreationUI(
                             isError = firstName.isBlank()
                         )
 
-                        Spacer(modifier = modifier.height(5.dp))
+                        Spacer(modifier = modifier.height(spacerHeight))
 
                         OutlinedTextField(
                             value = lastName,
@@ -267,7 +273,7 @@ fun ProfileCreationUI(
                     }
                     ProfilePictureEdit(picture, onPictureChange)
                 }
-                Spacer(modifier = modifier.height(5.dp))
+                Spacer(modifier = modifier.height(spacerHeight))
 
                 // Section and semester dropdown menus
                 DropDownListFunctionWrapper(
@@ -276,7 +282,7 @@ fun ProfileCreationUI(
                     selectedSec ?: "",
                     onSecChange
                 )
-                Spacer(modifier = modifier.height(5.dp))
+                Spacer(modifier = modifier.height(spacerHeight))
                 DropDownListFunctionWrapper(
                     semList,
                     R.string.profile_creation_semester,
@@ -284,7 +290,7 @@ fun ProfileCreationUI(
                     onSemChange
                 )
 
-                Spacer(modifier = modifier.height(10.dp))
+                Spacer(modifier = modifier.height(spacerHeight.times(2)))
 
                 // Tags
                 Text(
@@ -293,9 +299,9 @@ fun ProfileCreationUI(
                     fontWeight = FontWeight.Bold
                 )
 
-                Spacer(modifier = modifier.height(10.dp))
+                Spacer(modifier = modifier.height(spacerHeight.times(2)))
 
-                FlowRow(horizontalArrangement = Arrangement.spacedBy(5.dp)) {
+                FlowRow(horizontalArrangement = Arrangement.spacedBy(spaceBetweenTags)) {
                     for (tag in tagList) {
                         InputChipFun(tag.name) { tagDelete(tag) }
                     }
@@ -454,7 +460,10 @@ fun ProfilePictureEdit(picture: Bitmap?, onPictureChange: (newPicture: Bitmap?) 
             }
         }
     val pictureDisplaySize = 100.dp
-    Column(modifier = Modifier.padding(start = 5.dp)) {
+    val pictureStartPadding = 5.dp
+    val pictureAlpha = 0.5f
+    val deleteButtonOffset = 10.dp
+    Column(modifier = Modifier.padding(start = pictureStartPadding)) {
         Box {
             Image(
                 modifier =
@@ -475,7 +484,7 @@ fun ProfilePictureEdit(picture: Bitmap?, onPictureChange: (newPicture: Bitmap?) 
                         painterResource(R.drawable.echologoround)
                     },
                 contentDescription = "",
-                alpha = 0.5f
+                alpha = pictureAlpha
             )
             Icon(
                 imageVector = Icons.Outlined.Edit,
@@ -486,7 +495,7 @@ fun ProfilePictureEdit(picture: Bitmap?, onPictureChange: (newPicture: Bitmap?) 
         IconButton(
             modifier =
                 Modifier.align(Alignment.End)
-                    .offset(x = 10.dp, y = (-10).dp)
+                    .offset(x = deleteButtonOffset, y = -deleteButtonOffset)
                     .testTag("profile-picture-delete"),
             onClick = { onPictureChange(null) }
         ) {
@@ -529,6 +538,9 @@ fun PictureTransformer(
     val maxResolution = 500 // the picture is cropped to output a square
     var scale by remember { mutableStateOf(1f) }
     var offset by remember { mutableStateOf(Offset.Zero) }
+    val centerTextPadding = 5.dp
+    val centerTextCardOffset = 20.dp
+    val buttonsVerticalPadding = 15.dp
 
     // helper function to find out if the new transformation is inside the picture depending on
     // offset and scale
@@ -582,16 +594,16 @@ fun PictureTransformer(
                 style = Stroke(strokeSize)
             )
         }
-        Card(modifier = Modifier.align(Alignment.TopCenter).offset(y = 20.dp)) {
+        Card(modifier = Modifier.align(Alignment.TopCenter).offset(y = centerTextCardOffset)) {
             Text(
-                modifier = Modifier.padding(5.dp),
+                modifier = Modifier.padding(centerTextPadding),
                 text = stringResource(R.string.profile_creation_center_picture),
                 style = MaterialTheme.typography.titleLarge
             )
         }
         Row(
             modifier =
-                Modifier.align(Alignment.BottomEnd).padding(horizontal = 0.dp, vertical = 15.dp)
+                Modifier.align(Alignment.BottomEnd).padding(vertical = buttonsVerticalPadding)
         ) {
             val button_Padding = 5.dp
             FilledTonalButton(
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/github/swent/echo/compose/authentication/CreateProfile.kt
```diff
@@ -194,31 +228,45 @@ fun ProfileCreationUI(
...
-                    value = lastName,
-                    onValueChange = { onLastNameChange(it) },
-                    modifier = modifier.fillMaxWidth().testTag("LastName"),
-                    label = {
-                        Text(text = stringResource(id = R.string.profile_creation_last_name))
-                    },
-                    singleLine = true,
-                    isError = lastName.isBlank()
-                )
+                        Spacer(modifier = modifier.height(5.dp))
```
ntissieres, line 250: The height of the spacer can be set into a val at the beginning of the file, while it is use multiple times

app/src/main/java/com/github/swent/echo/compose/authentication/CreateProfile.kt
```diff
@@ -376,3 +424,197 @@ fun InputChipFun(
...
+                    BitmapFactory.decodeStream(localContext.contentResolver.openInputStream(uri))
+                if (rawPicture == null) {
+                    Log.w("photo picker", "cannot read the file")
+                } else {
+                    showPictureDialog = true
+                }
+            }
+        }
+    val pictureDisplaySize = 100.dp
+    Column(modifier = Modifier.padding(start = 5.dp)) {
```
ntissieres, line 457: Again, try to avoid magic values in padding

app/src/main/java/com/github/swent/echo/compose/authentication/CreateProfile.kt
```diff
@@ -376,3 +424,197 @@ fun InputChipFun(
...
+                modifier =
+                    Modifier.size(pictureDisplaySize).clip(CircleShape).clickable {
+                        pickPhotoActivity.launch(
+                            PickVisualMediaRequest(
+                                ActivityResultContracts.PickVisualMedia.ImageOnly
+                            )
+                        )
+                    },
+                painter =
+                    if (picture != null) {
```
ntissieres, line 472: Nice idea to use the echo logo if no profile picture !

app/src/main/java/com/github/swent/echo/compose/authentication/CreateProfile.kt
```diff
@@ -376,3 +424,197 @@ fun InputChipFun(
...
+                        BitmapPainter(picture!!.asImageBitmap())
+                    } else {
+                        painterResource(R.drawable.echologoround)
+                    },
+                contentDescription = "",
+                alpha = 0.5f
```
ntissieres, line 478: magic value

app/src/main/java/com/github/swent/echo/compose/authentication/CreateProfile.kt
```diff
@@ -376,3 +424,197 @@ fun InputChipFun(
...
+            )
+            Icon(
+                imageVector = Icons.Outlined.Edit,
+                contentDescription = stringResource(R.string.profile_creation_edit_picture),
+                modifier = Modifier.align(Alignment.Center)
+            )
+        }
+        IconButton(
+            modifier = Modifier.align(Alignment.End).offset(x = 10.dp, y = (-10).dp),
```
ntissieres: magic values

app/src/main/java/com/github/swent/echo/compose/authentication/CreateProfile.kt
```diff
@@ -376,3 +424,197 @@ fun InputChipFun(
...
+        )
+        val strokeSize = 10f
+        Canvas(modifier = Modifier.fillMaxSize()) {
+            drawCircle(
+                color = Color.Black,
+                radius = (circleRadius - strokeSize).toFloat(),
+                style = Stroke(strokeSize)
+            )
+        }
+        Card(modifier = Modifier.align(Alignment.TopCenter).offset(y = 20.dp)) {
```
ntissieres, line 585: magic value

app/src/main/java/com/github/swent/echo/compose/authentication/CreateProfile.kt
```diff
@@ -376,3 +424,197 @@ fun InputChipFun(
...
+            Text(
+                modifier = Modifier.padding(5.dp),
```
ntissieres, line 587: magic value

app/src/main/java/com/github/swent/echo/compose/authentication/CreateProfile.kt
```diff
@@ -376,3 +424,197 @@ fun InputChipFun(
...
+                text = stringResource(R.string.profile_creation_center_picture),
+                style = MaterialTheme.typography.titleLarge
+            )
+        }
+        Row(
+            modifier =
+                Modifier.align(Alignment.BottomEnd).padding(horizontal = 0.dp, vertical = 15.dp)
```
ntissieres, line 594: magic values

app/src/main/java/com/github/swent/echo/viewmodels/HomeScreenViewModel.kt
```diff
@@ -161,6 +165,13 @@ constructor(
             followedTagFilter = getTagsAndSubTags(_followedTags.value).toList()
             sectionTags = repository.getSubTags(sectionTagId)
             semesterTags = repository.getSubTags(semesterTagId)
+            val pictureByteArray = repository.getUserProfilePicture(userId)
+            _profilePicture.value =
+                if (pictureByteArray != null) {
+                    BitmapFactory.decodeByteArray(pictureByteArray, 0, pictureByteArray.size)
+                } else {
+                    null
```
srsingh04, line 173: Is the else block necessary? The default value is set to null, so you could safely get rid of the else block.

app/src/main/java/com/github/swent/echo/viewmodels/authentication/CreateProfileViewModel.kt
```diff
@@ -94,6 +100,17 @@ constructor(
...
+                    val pictureByteArray = repository.getUserProfilePicture(userId)
+                    _picture.value =
+                        if (pictureByteArray != null) {
+                            BitmapFactory.decodeByteArray(
+                                pictureByteArray,
+                                0,
+                                pictureByteArray.size
+                            )
+                        } else {
+                            null
```
srsingh04, line 112: Similarly, the else block is not required here because the default value is null already.

ntissieres (changes requested):
Truly impressive code ! You realy manage to make the feature of adding profiles pictures very handy for the user, nice work ! Still, there is some minor changes required :
- A lot of magic values to put into val in order to make the maintenance easier in the case we want to change things
- The profile picture in the hamburger menu is a bit overlapping the switch dark/light mode. It could be good to replace the "close hamburger menu" button (the cross) which is not necessary anymore with the light/dark mode button.
- The center picture feature is good, but instead of using the circle to indicate the center of the picture, it would be better to enlarge it to show instead what part of the picture will be selected

Otherwise, I tried to change my profile picture and it works great ;)

app/src/main/java/com/github/swent/echo/compose/authentication/CreateProfile.kt
```diff
@@ -376,3 +424,197 @@ fun InputChipFun(
...
+    }
+}
+
+/**
+ * Picture edition dialog which allow the user to center the picture
+ *
+ * @param picture: the picture to center
+ * @param onConfirm: callback called when the picture is centered
+ * @param onCancel: callback when the action is canceled
+ */
```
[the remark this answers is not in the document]
unglazedstamp: Yes I really wanted to do something similar to what most messaging app do with the profile picture, it's not as good as these apps but it's clearly usable.

app/src/main/java/com/github/swent/echo/compose/authentication/CreateProfile.kt
```diff
@@ -376,3 +424,209 @@ fun InputChipFun(
...
+    var scale by remember { mutableStateOf(1f) }
+    var offset by remember { mutableStateOf(Offset.Zero) }
+
+    // helper function to find out if the new transformation is inside the picture depending on
+    // offset and scale
+    fun isTransformInside(offs: Offset, sca: Float): Boolean {
+        return offs.x > -picture.width * sca / 2f + circleRadius &&
+            offs.y > -picture.height * sca / 2f + circleRadius &&
+            offs.x < picture.width * sca / 2f - circleRadius &&
+            offs.y < picture.height * sca / 2f - circleRadius &&
```
srsingh04, line 539: you could modularize the return statement by defining each set of conditions as variables and then using them in the return statement. This would also allow someone in the future to easily edit the conditions without much hassle.

app/src/main/java/com/github/swent/echo/compose/authentication/CreateProfile.kt
```diff
@@ -376,3 +424,209 @@ fun InputChipFun(
...
+
+    val state = rememberTransformableState { zoomChange, offsetChange, _ ->
+        val newScale = scale * zoomChange
+        val newOffset = offset + offsetChange
+        scale =
+            if (isTransformInside(offset, newScale)) {
+                newScale
+            } else {
+                scale
+            }
```
srsingh04, line 552: Again, I think the else block can be safely deleted because the scale variable just keeps its value when the condition is not respected.

app/src/main/java/com/github/swent/echo/compose/authentication/CreateProfile.kt
```diff
@@ -376,3 +424,209 @@ fun InputChipFun(
...
+
+        offset =
+            if (isTransformInside(newOffset, scale)) {
+                newOffset
+            } else {
+                offset
```
srsingh04, line 558: idem, else block can be safely deleted.

app/src/test/java/com/github/swent/echo/compose/authentication/CreateProfileTest.kt
```diff
@@ -62,4 +83,80 @@ class CreateProfileTest {
...
+                onCancel = {}
+            )
+        }
+        val image = composeTestRule.onNodeWithTag("profile-picture-image")
+        image.assertIsDisplayed()
+        image.performTouchInput { this.swipe(Offset.Zero, Offset(50f, 50f)) }
+        image.assertPositionInRootIsEqualTo(15.dp, 15.dp)
+    }
+
+    // this test needs to be run there because of the Bitmap class which require android
```
srsingh04, line 144: Knitpick: the comment is a bit confusing: what exactly do you mean by 'test needs to be run there'? run where? Maybe I am just confused for nothing here...

srsingh04 (changes requested):
Nice code, thank you for working on the feature. Some minor changes to be done and then it should be good to go. 
Overall good work.

app/src/main/java/com/github/swent/echo/compose/authentication/CreateProfile.kt
```diff
@@ -376,3 +424,209 @@ fun InputChipFun(
...
+ *
+ * @param picture: the picture to display
+ * @param onPictureChange: callback called when the picture change
+ */
+@Composable
+fun ProfilePictureEdit(picture: Bitmap?, onPictureChange: (newPicture: Bitmap?) -> Unit) {
+    var showPictureDialog by remember { mutableStateOf(false) }
+    val localContext = LocalContext.current
+    if (!isPhotoPickerAvailable(localContext)) {
+        Log.e("CreateProfile", "Photo picker not available")
```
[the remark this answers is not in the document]
unglazedstamp: If the photo picker isn't available, android uses a fallback, so the feature is still relevant, however our application isn't supposed to be used on these old device and some error could arise (e.g. select a file which isn't a picture) so I log an error

app/src/main/java/com/github/swent/echo/compose/authentication/CreateProfile.kt
```diff
@@ -376,3 +424,209 @@ fun InputChipFun(
...
+    }
+    var rawPicture by remember { mutableStateOf<Bitmap?>(null) }
+    val pickPhotoActivity =
+        rememberLauncherForActivityResult(ActivityResultContracts.PickVisualMedia()) { uri ->
+            if (uri == null) {
+                Log.w("photo picker", "photo not found")
+            } else {
```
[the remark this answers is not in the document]
unglazedstamp: I believe that it's better for maintainability to have a clear if else statement, furthermore the else statement allow for smart casting which is good for code readability and safety.

app/src/test/java/com/github/swent/echo/compose/authentication/CreateProfileTest.kt
unglazedstamp: ok I will remove this comment

srsingh04 (dismissed):
Great! Thank you for the nice feature. Ready to be merged from my side.

app/src/main/java/com/github/swent/echo/compose/authentication/CreateProfile.kt
```diff
@@ -376,3 +430,210 @@ fun InputChipFun(
...
+ * @param onConfirm: callback called when the picture is centered
+ * @param onCancel: callback when the action is canceled
+ */
+@Composable
+fun PictureTransformer(
+    picture: Bitmap,
+    onConfirm: (picture: Bitmap) -> Unit,
+    onCancel: () -> Unit
+) {
+    val circleRadius = min(picture.width, picture.height) * 3 / 8
```
ntissieres, line 536: I don't think it's a good idea to set the circle radius according the image size. I tried it with different image, and some mystical things happens if the image is not the same size as the screen. I think it would be preferable to use a fix size

unglazedstamp: The problem here is that if the picture is smaller than the fixed size then the app crashes

PR #361, opened 2024-05-28
👌 IMPROVE: make repository robust in treating network failures

### LINKED ISSUE #360: Make the repository more robust
retry on network issues and avoid failures

violoncelloCH:
closes #360

### DIFF
--- a/app/src/main/java/com/github/swent/echo/data/repository/RepositoryImpl.kt
+++ b/app/src/main/java/com/github/swent/echo/data/repository/RepositoryImpl.kt
@@ -2,6 +2,7 @@ package com.github.swent.echo.data.repository
 
 import com.github.swent.echo.connectivity.NetworkService
 import com.github.swent.echo.data.model.Association
+import com.github.swent.echo.data.model.DataModel
 import com.github.swent.echo.data.model.Event
 import com.github.swent.echo.data.model.Tag
 import com.github.swent.echo.data.model.UserProfile
@@ -40,14 +41,14 @@ class RepositoryImpl(
         var tags_last_cached_subs: MutableMap<String, Long> = mutableMapOf()
     }
 
-    private suspend fun <DataObject> getObject(
+    private suspend fun <T : DataModel> getObject(
         objectId: String,
         objectTTL: Long,
-        getLocal: suspend (String, Long) -> DataObject?,
-        getRemote: suspend (String) -> DataObject?,
+        getLocal: suspend (String, Long) -> T?,
+        getRemote: suspend (String) -> T?,
         deleteLocal: suspend (String) -> Unit,
-        setLocal: suspend (DataObject) -> Unit
-    ): DataObject? {
+        setLocal: suspend (T) -> Unit
+    ): T? {
         if (isOffline()) {
             return getLocal(objectId, FETCH_ALL)
         }
@@ -69,13 +70,13 @@ class RepositoryImpl(
         return localResult
     }
 
-    private suspend fun <DataObject> getObjects(
+    private suspend fun <T : DataModel> getObjects(
         objectIds: List<String>,
         objectTTL: Long,
-        getLocal: suspend (List<String>, Long) -> List<DataObject>,
-        getRemote: suspend (List<String>) -> List<DataObject>,
-        setLocal: suspend (List<DataObject>) -> Unit
-    ): List<DataObject> {
+        getLocal: suspend (List<String>, Long) -> List<T>,
+        getRemote: suspend (List<String>) -> List<T>,
+        setLocal: suspend (List<T>) -> Unit
+    ): List<T> {
         if (isOffline()) {
             return getLocal(objectIds, FETCH_ALL)
         }
@@ -94,14 +95,14 @@ class RepositoryImpl(
         return localResult
     }
 
-    private suspend fun <DataObject> getAllObjects(
+    private suspend fun <T : DataModel> getAllObjects(
         objectsLastCachedAll: MutableList<Long>,
         objectTTL: Long,
-        getAllLocal: suspend (Long) -> List<DataObject>,
-        getAllNotInRemote: suspend (List<String>) -> List<DataObject>,
+        getAllLocal: suspend (Long) -> List<T>,
+        getAllNotInRemote: suspend (List<String>) -> List<T>,
         deleteAllNotInLocal: suspend (List<String>) -> Unit,
-        setLocal: suspend (List<DataObject>) -> Unit
-    ): List<DataObject> {
+        setLocal: suspend (List<T>) -> Unit
+    ): List<T> {
         if (isOffline() || notExpired(objectsLastCachedAll[0], objectTTL)) {
             return getAllLocal(FETCH_ALL)
         }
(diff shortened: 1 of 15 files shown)


app/src/main/java/com/github/swent/echo/data/repository/RepositoryImpl.kt
```diff
@@ -45,8 +46,17 @@ class RepositoryImpl(
         }
         val localResult = localDataSource.getAssociation(associationId, ASSOCIATION_CACHE_TTL)
         if (localResult == null) {
-            val remoteResult = remoteDataSource.getAssociation(associationId)
-            if (remoteResult != null) localDataSource.setAssociation(remoteResult)
+            val remoteResult =
+                try {
+                    remoteDataSource.getAssociation(associationId)
+                } catch (e: HttpRequestException) {
+                    return localDataSource.getAssociation(associationId, FETCH_ALL)
```
octogradiste, line 53: As you are using the try catch block as an expression you can drop the `return` key word (the last statement is automatically returned). You can also do this in all try catch blocks below.

app/src/main/java/com/github/swent/echo/data/supabase/SupabaseDataSource.kt
```diff
@@ -63,112 +70,271 @@ class SupabaseDataSource(private val supabase: SupabaseClient) : RemoteDataSourc
...
-            null
+        } catch (e: Exception) {
+            when (e) {
+                is BadRequestRestException -> null
+                is HttpRequestException -> {
+                    if (maxRetriesCount == 0u) throw e
+                    delay(RETRY_DELAY_MILLI)
+                    return getAssociation(associationId, maxRetriesCount - 1u)
+                }
+                else -> throw e
```
octogradiste, line 81: You could write a function which wraps the supabase call and reuse it everywhere. It could look like this:
```kotlin
/**
 * Wrapper for Supabase requests that retries the request if it fails due to a
 * [HttpRequestException]. If the request fails after [maxRetriesCount] retries, it throws the
 * exception. In case of an [NoSuchElementException], it returns null.
 *
 * @param maxRetriesCount: the number of times the request will be retried
 * @param request: the request to be executed
 */
private suspend fun <T> supabaseRequestWrapper(
    maxRetriesCount: UInt,
    request: suspend () -> T,
): T? {
    return try {
        request()
    } catch (e: NoSuchElementException) {
        null
    } catch (e: HttpRequestException) {
        if (maxRetriesCount == 0u) throw e
        delay(RETRY_DELAY_MILLI)
        return supabaseRequestWrapper(maxRetriesCount - 1u, request)
    }
}
```

And you could use it like this:
```kotlin
override suspend fun getAssociation(
    associationId: String,
    maxRetriesCount: UInt
): Association? = supabaseRequestWrapper(maxRetriesCount) {
    supabase
        .from("associations")
        .select(Columns.raw(QUERY_ASSOCIATION)) {
            filter { eq("association_id", associationId) }
        }
        .decodeSingle<AssociationSupabase>()
        .toAssociation()
}
```

Note that you would need to also write a wrapper fur function which return a list.

octogradiste (changes requested):
Again massive work! Thank you so much! Consider creating a wrapper function for the supabase data source. This would avoid some duplication and make it cleaner :). Ottherwise, looks really good to me!

app/src/main/java/com/github/swent/echo/data/repository/RepositoryImpl.kt
violoncelloCH: Thanks for you feedback. 
Taking a close look on this, I realize that no, this is not the case. The return is intentional here to exit the function.
If you check the line above the try (currently 49), there is a variable assignment done. Based on that variable assignment, further actions are taken. However in case of the exception the result from line 53 needs to be returned directly as otherwise it would be written back to the local database unnecessarily.

app/src/main/java/com/github/swent/echo/data/supabase/SupabaseDataSource.kt
violoncelloCH: Thanks! Yes absolutely. I've been thinking about generalizing these. Actually there are quite a few things which are similar for each type of methods (single getters, getAlls, deletes, etc). However there are still some subtle differences for some of them and I'm unsure up to which point generalization makes sense. Even for the wrapper you're suggesting, there are actually some functions for which I check for different Exceptions.
I'll look into this and see if I can build something reasonable without breaking the tests :D

violoncelloCH: Okay, so I heavily reworked this. Based on your suggestion I did a wrapper function for the retry and exception handling in the SupabaseDataSource. In a similar way, i reworked the code from the RepositoryImpl to remove redundancies as much as possible. (The changes are in the last two commits)

app/src/main/java/com/github/swent/echo/data/repository/RepositoryImpl.kt
```diff
@@ -288,20 +408,32 @@ class RepositoryImpl(
...
-        "Storing/Deleting a " + objectTryingToStore + " while the App is offline is not supported"
-    )
+private fun <DataObject> DataObject.getId(): String =
+    when (this) {
+        is Association -> this.associationId
+        is Event -> this.eventId
+        is Tag -> this.tagId
+        is UserProfile -> this.userId
+        else -> ""
+    }
```
octogradiste, line 439: Instead of using a generic type `DataObject`, I'd suggest you use a sealed class `DataModel` for all our data model, this would make it typesafe and avoid the `else` branch.

octogradiste (changes requested):
Thanks! The code already looks much cleaner. Just some minor changes to make it even more clean ;)

app/src/main/java/com/github/swent/echo/data/supabase/SupabaseDataSource.kt
```diff
@@ -47,145 +50,250 @@ class SupabaseDataSource(private val supabase: SupabaseClient) : RemoteDataSourc
...
-                filter { isIn("association_id", associations) }
-            }
-            .decodeList<AssociationSupabase>()
-            .toAssociations()
-    }
+    override suspend fun getAssociations(
+        associations: List<String>,
+        maxRetriesCount: UInt
+    ): List<Association> =
+        supabaseRequestExceptionHandlerAndRetryWrapperS(maxRetriesCount) {
```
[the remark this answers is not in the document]
violoncelloCH: ohh yes, good Idea! I was too tired yesterday evening to come up with a cleaner solution :sweat_smile: 
thanks!

app/src/main/java/com/github/swent/echo/data/repository/RepositoryImpl.kt
violoncelloCH: I tried creating a supertype for the dataclasses, but somehow dataclasses can't inherit in Kotlin. I was not aware that I could use sealed classes for this. Indeed it is the best solution for this! Thank you very much!

octogradiste (approved):
Thank you so much! Now the code is really clean!

PR #363, opened 2024-05-29
Update/rework search bottom sheet

ntissieres:
Updating the UI

### DIFF
--- a/app/src/main/java/com/github/swent/echo/compose/components/HomeScreen.kt
+++ b/app/src/main/java/com/github/swent/echo/compose/components/HomeScreen.kt
@@ -119,6 +119,8 @@ private fun Content(
     // search mode for displaying events
     val searchMode by homeScreenViewModel.searchMode.collectAsState()
 
+    val initialPage by homeScreenViewModel.initialPage.collectAsState()
+
     // online status for disabling the buttons
     val isOnline by homeScreenViewModel.isOnline.collectAsState()
 
@@ -188,7 +190,9 @@ private fun Content(
                 fullCallback = homeScreenViewModel::onFullCheckedSwitch,
                 sortByCallback = homeScreenViewModel::onSortByChanged,
                 resetFiltersCallback = homeScreenViewModel::resetFiltersContainer,
-                timeFilterCallback = homeScreenViewModel::onDateFilterChanged
+                timeFilterCallback = homeScreenViewModel::onDateFilterChanged,
+                initialPage = initialPage,
+                mode = mode
             )
             // {navActions.navigateTo(Routes.SearchScreen)}) <- when we make a whole screen for
             // the search menu
--- a/app/src/main/java/com/github/swent/echo/compose/components/SearchMenuSheet.kt
+++ b/app/src/main/java/com/github/swent/echo/compose/components/SearchMenuSheet.kt
@@ -1,33 +1,21 @@
 package com.github.swent.echo.compose.components
 
 import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
-import androidx.compose.foundation.layout.Spacer
-import androidx.compose.foundation.layout.absoluteOffset
-import androidx.compose.foundation.layout.fillMaxHeight
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
-import androidx.compose.foundation.layout.width
-import androidx.compose.material.icons.Icons
-import androidx.compose.material.icons.filled.Settings
-import androidx.compose.material.icons.filled.ShoppingCart
 import androidx.compose.material3.Button
 import androidx.compose.material3.ExperimentalMaterial3Api
-import androidx.compose.material3.Icon
 import androidx.compose.material3.ModalBottomSheet
-import androidx.compose.material3.OutlinedTextField
 import androidx.compose.material3.SheetValue
 import androidx.compose.material3.Text
 import androidx.compose.material3.rememberModalBottomSheetState
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.MutableState
-import androidx.compose.runtime.mutableStateOf
-import androidx.compose.runtime.remember
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
-import androidx.compose.ui.graphics.vector.ImageVector
 import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.unit.dp
@@ -36,7 +24,7 @@ import com.github.swent.echo.R
 import com.github.swent.echo.compose.components.searchmenu.FiltersContainer
 import com.github.swent.echo.compose.components.searchmenu.SearchMenuDiscover
 import com.github.swent.echo.compose.components.searchmenu.SearchMenuFilters
-import com.github.swent.echo.compose.components.searchmenu.SortBy
+import com.github.swent.echo.viewmodels.MapOrListMode
 import com.github.swent.echo.viewmodels.tag.TagViewModel
 
 @OptIn(ExperimentalMaterial3Api::class)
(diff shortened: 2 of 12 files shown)


app/src/main/java/com/github/swent/echo/compose/components/SearchMenuSheet.kt
```diff
@@ -52,129 +40,84 @@ fun SearchMenuSheet(
...
             }
             true
         }
+    val bottomSheetHeight = 500.dp
+    val paddingValues = 5.dp
+    val pageHeight = 385.dp
+    val paddingBottomResetFilters = 10.dp
     ModalBottomSheet(
-        modifier = Modifier.fillMaxWidth().fillMaxHeight(0.5f).testTag("search_menu_sheet"),
+        modifier = Modifier.fillMaxWidth().height(bottomSheetHeight).testTag("search_menu_sheet"),
```
Greenade, line 67: this is actually not recommended, fillMaxHeight(0.5f) does so that the sheet takes half the phone size, the way you did things makes it a fixed size. Unless you had a reason to do this I suggest keeping fillMaxHeight

app/src/main/java/com/github/swent/echo/compose/components/SearchMenuSheet.kt
```diff
@@ -52,129 +40,84 @@ fun SearchMenuSheet(
...
-                    SearchMenuDiscover(searchEntryCallback, tagViewModel)
-                }
+                // Sheet content
+                SearchBar(
+                    stringResource(R.string.search_menu_sheet_search_interests),
+                    filters.searchEntry,
+                    searchEntryCallback
+                )
+                // Display filters or discover according to the selected mode
+                Pager(
```
Greenade, line 85: nice addition ! makes more sense than the button previously there

app/src/main/java/com/github/swent/echo/compose/components/searchmenu/SearchMenuFilters.kt
```diff
@@ -114,40 +111,54 @@ fun SearchMenuFilters(
                 fullCallback()
             }
         )
+    val followedAssociations = listOf("1", "2", "3", "4", "5")
```
Greenade, line 114: will need to be replaced by an argument that propagates from the viewmodel, you should already add it to the signature. Same for selectedAssociation and the callback

app/src/main/java/com/github/swent/echo/compose/components/searchmenu/SearchMenuFilters.kt
```diff
@@ -114,40 +111,54 @@ fun SearchMenuFilters(
...
-            )
-        }
-        Row(modifier = Modifier.align(Alignment.TopCenter).absoluteOffset(y = 170.dp)) {
-            DateFilter(filters, timeFilterCallback)
-        }
+    Column(
+        modifier = Modifier.fillMaxSize().testTag("search_menu_filters_content"),
+        horizontalAlignment = Alignment.CenterHorizontally
+    ) {
+        TwoBoxesDisplayer(
```
Greenade, line 123: nice way of doing this, very scalable and looking good 👍

app/src/main/java/com/github/swent/echo/compose/components/searchmenu/SearchMenuFilters.kt
```diff
@@ -293,6 +281,17 @@ fun DateFilter(filters: FiltersContainer, timeSliderCallback: (Float, Float) ->
     }
 }
 
+@Composable
+fun TwoBoxesDisplayer(firstBox: @Composable () -> Unit, secondBox: (@Composable () -> Unit)?) {
```
Greenade, line 285: nice addition, makes the code very readable and looks nice

app/src/main/java/com/github/swent/echo/viewmodels/HomeScreenViewModel.kt
```diff
@@ -99,7 +107,7 @@ constructor(
             fullChecked = false,
             from = 0f,
             to = 14f,
-            sortBy = SortBy.NONE
+            sortBy = null
```
Greenade, line 110: I think it doesn't really make sense that one can set the sortby to none, this would mean the list is not sorted when in reality it always is sorted the way we did it. This should probably be set to sortby.DATE_ASC to make sense in my opinion

app/src/main/java/com/github/swent/echo/viewmodels/HomeScreenViewModel.kt
```diff
@@ -130,6 +138,8 @@ constructor(
     // Flow to observe the search mode
     private var _searchMode = MutableStateFlow(false)
     val searchMode = _searchMode.asStateFlow()
+    private val _initialPage = MutableStateFlow(0)
```
Greenade, line 141: please add a comment to explain what initial page does

Greenade (changes requested):
Thank you for your work ! This is a huge PR, it's looking good ! I added some comments about sorting and comments but other than that it's looking great !

app/src/main/java/com/github/swent/echo/compose/components/searchmenu/SearchMenuFilters.kt
Greenade: I will implement the changes to the viewmodel obviously as it is my task, but you should already make the signature consistent

app/src/main/java/com/github/swent/echo/compose/components/HomeScreen.kt
```diff
@@ -119,6 +119,8 @@ private fun Content(
     // search mode for displaying events
     val searchMode by homeScreenViewModel.searchMode.collectAsState()
```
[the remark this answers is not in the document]
ntissieres: The comment will be added in the next pr where i will comment all my code. But it avoids using discover mode, closing the bottom sheet to look at the event, and reopening the bottom sheet which would display filters instead of discover. Concretely, it remembers the last page you openned

app/src/main/java/com/github/swent/echo/compose/components/SearchMenuSheet.kt
ntissieres: The reason is that each phone have a different size. If we have a fix size, we are guaranteed that all elements are always displayed the same and that there is no missing components on little phones

app/src/main/java/com/github/swent/echo/compose/components/searchmenu/SearchMenuFilters.kt
ntissieres: I can do it if you want, i thought it was part of the integration with the viewmodel since you just have to propagate those values and call .collectAsState.value in Homescreen

app/src/main/java/com/github/swent/echo/viewmodels/HomeScreenViewModel.kt
ntissieres: They are not sorted for now (check the events) but I agree it could be set to DATE_ASC. It's more a user convenience to see a dropDown named SortBy instead of DATE_ASC. I don't know if it's a big problem but I can change if you think it's necessary

app/src/main/java/com/github/swent/echo/compose/components/searchmenu/SearchMenuFilters.kt
Greenade: The thing is it makes more sense to have the correct signature from the start, but I can still do it in my next task that's fine

app/src/main/java/com/github/swent/echo/viewmodels/HomeScreenViewModel.kt
Greenade: Are they not sorted by date automatically ? They should because that's how I did things

ntissieres: I corrected everything you said above, thanks for the useful feedback !!

Greenade (dismissed):
Thank you for the changes 🙏 it's ready to merge once you update the tests

Greenade (approved):
APPROVED 💯

PR #364, opened 2024-05-29
Feature/adding association new attributes

ntissieres:
Incorporation of the new attributes in the UI : 
- AssociationDetails : Show all related tags in the description page and add a hyperlink to the association url if it has one
- AssociationViewModel : We can now search association by tags

### DIFF
--- a/app/src/main/java/com/github/swent/echo/compose/association/AssociationDetails.kt
+++ b/app/src/main/java/com/github/swent/echo/compose/association/AssociationDetails.kt
@@ -52,7 +52,7 @@ fun AssociationDetails(
     val followWidth = 150.dp
     val followHeight = 40.dp
     val followSpaceInside = 5.dp
-    val verticalSpace = 20.dp
+    val verticalSpace = 12.dp
     Column(
         modifier = Modifier.fillMaxSize().padding(paddingValues).testTag("association_details")
     ) {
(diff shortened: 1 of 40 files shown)


app/src/main/java/com/github/swent/echo/compose/association/AssociationDetails.kt
```diff
@@ -90,8 +99,46 @@ fun AssociationDetails(
...
+            }
+        }
+        Text(association.description)
+        if (association.url != null) {
+            Spacer(modifier = Modifier.height(verticalSpace))
+            Text(
+                stringResource(R.string.association_details_contact),
+                style = MaterialTheme.typography.titleSmall
+            )
+            Spacer(modifier = Modifier.height(verticalSpace))
```
Greenade, line 138: This spacer is a bit too big in my opinion

app/src/main/res/values/strings.xml
```diff
@@ -130,5 +130,6 @@
     <string name="authentication_error_email_is_required">Email is required</string>
     <string name="authentication_password_too_short">Password must be at least 8 characters</string>
     <string name="authentication_error_passwords_do_not_match">Passwords do not match</string>
+    <string name="association_details_contact">Contact:</string>
```
Greenade, line 133: you should add a space between contact and the column (Contact :)

Greenade (changes requested):
Thank you for the work on associations ! I asked for minor changes but otherwise it's looking amazing ! 💯

Greenade (approved):
Thank you for the changes ! It's perfect now !

PR #365, opened 2024-05-29
translation to french 

srsingh04:
Now, the user can enjoy the application in french as well, when their system settings are set to french language. 
Along with that, some paddings and other things had to be changed in the UI to better accomodate the user interface with the french language.


alejandrocalles (changes requested):
Hey, thanks for all this work! I left some small comments on the grammar, but other than that, the changes looks great.

There's on small bug though, which I'm not sure is very relevant: when I open the app in English, then I go to settings and change the language to French, and then come back, the app crashes. Maybe this is just my phone, but it would be great you could look into this.

In any case, changing the language while the app is running is a very niche user flow, so it may not be too important, but it should be doable.

For instance, when I do the same thing with Chrome, it doesn't crash, but asks whether I want to change the app's language to the device's language.

alejandrocalles:
Thanks! Approving.

alejandrocalles (approved):

PR #367, opened 2024-05-30
Fix app crash when switching theme

### LINKED ISSUE #333: App crashes when switching the color scheme

octogradiste:
The problem was that we kept the same instance of the `mapView` across configuration changes. Therefore, when switching the theme, the `mapView` already had a parent view which causes an `IllegalStateException`. To fix this, I recreate the `mapView` at every configuration change. This has the side effect of re-centering the map at every configuration change. To avoid this, I save the camera position every time it moves, so I can restore the previous camera position when the map is recreated. Closes #333.

### DIFF
--- a/app/src/main/java/com/github/swent/echo/compose/map/MapLibreMapViewProvider.kt
+++ b/app/src/main/java/com/github/swent/echo/compose/map/MapLibreMapViewProvider.kt
@@ -109,9 +109,7 @@ class MapLibreMapViewProvider : IMapViewProvider<MapView> {
                 }
 
                 // Add listener to update the last camera position so we can restore it later
-                map.addOnCameraMoveListener {
-                    lastCameraPosition = CameraPosition.Builder(map.cameraPosition).build()
-                }
+                map.addOnCameraMoveListener { lastCameraPosition = map.cameraPosition }
 
                 // Call the on create callback once the map is configured
                 onCreate()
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/github/swent/echo/compose/map/MapLibreMapViewProvider.kt
```diff
@@ -79,36 +88,37 @@ class MapLibreMapViewProvider : IMapViewProvider<MapView> {
...
+
+                // Add listener to handle long press events
                 map.addOnMapLongClickListener {
                     onLongPress(it)
                     true
                 }
+
+                // Add listener to update the last camera position so we can restore it later
+                map.addOnCameraMoveListener {
+                    lastCameraPosition = CameraPosition.Builder(map.cameraPosition).build()
```
alejandrocalles, line 113: I believe it could be more readable (and optimal) to assign the map's camera position directly instead of building a new one each time the map moves.

alejandrocalles (changes requested):
Amazing, thanks for fixing this!  Good job on finding the source of the issue :)

I've made a small request for change, but other than that the code looks really clean.

One thing we could consider, in the future, is a way to only re-create the map on configuration changes, as it seems mildly inefficient to re-create the map every time we navigate to it (or every time we switch from list mode to map mode).
However, for the moment, the app seems to be running as smoothly as before, so it may not even be an issue.

alejandrocalles:
LGTM. Thanks :)

alejandrocalles (approved):

PR #368, opened 2024-05-30
Feature/filter by associations

Greenade:
make it possible to filter by an association you follow through the dropdown menu in the search sheet

### DIFF
--- a/app/src/main/java/com/github/swent/echo/compose/components/searchmenu/SearchMenuFilters.kt
+++ b/app/src/main/java/com/github/swent/echo/compose/components/searchmenu/SearchMenuFilters.kt
@@ -52,7 +52,7 @@ fun SearchMenuFilters(
     fullCallback: () -> Unit, // Callback function when the full filter is toggled
     sortByCallback: (Int) -> Unit, // Callback function when the sort by filter is changed
     timeFilterCallback: (Float, Float) -> Unit, // Callback function when the time filter is changed
-    mode: MapOrListMode // The current display mode (map or list)
+    mode: MapOrListMode, // The current display mode (map or list)
     followedAssociations: List<String>,
     selectedAssociation: Int,
     associationCallback: (Int) -> Unit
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/github/swent/echo/viewmodels/HomeScreenViewModel.kt
```diff
@@ -376,78 +399,86 @@ constructor(
...
+                        // filter by semester
                         !_filtersContainer.value.classChecked ||
                             _semester.value == "" ||
                             event.tags.any { tag ->
                                 tag.name.lowercase() == _semester.value.lowercase()
                             }
                     }
+                    .filter { event ->
+                        // filter by association
+                        _selectedAssociation.value == -1 ||
```
srsingh04, line 463: Knitpick: -1 can be defined as a private constant and then used here instead of directly using the value.

app/src/main/java/com/github/swent/echo/compose/components/searchmenu/SearchMenuFilters.kt
```diff
@@ -134,7 +134,7 @@ fun SearchMenuFilters(
                     Dropdown(
                         "Sort By",
                         SortBy.entries.map { stringResource(it.stringKey) },
-                        filters.sortBy?.ordinal ?: -1,
+                        filters.sortBy.ordinal,
```
srsingh04, line 137: Question: Any particular reason as you to why you removed the null pointer check?

app/src/main/java/com/github/swent/echo/viewmodels/HomeScreenViewModel.kt
```diff
@@ -310,7 +332,8 @@ constructor(
      * the default one, the search mode is on.
      */
     private fun updateSearchMode() {
-        _searchMode.value = _filtersContainer.value != defaultFiltersContainer
+        _searchMode.value =
+            _filtersContainer.value != defaultFiltersContainer || _selectedAssociation.value != -1
```
srsingh04, line 336: Again, the magic value -1 can be used in as a constant.

srsingh04 (approved):
Thank you for making the changes. Have added some minor suggestions for the code; ready to be merged nonetheless.

app/src/main/java/com/github/swent/echo/compose/components/searchmenu/SearchMenuFilters.kt
Greenade: Nathan changed it so it's not nullable anymore, so there's no need for a null check

srsingh04 (commented):
Nice! Thank you for the changes and explanations!

srsingh04 (approved):
Ready to be merged

PR #369, opened 2024-05-30
Fix/comment code nathan

ntissieres:


Greenade (approved):
Thanks for the comments ! The files are much clearer now !

PR #372, opened 2024-05-31
Connect the modify event callback

octogradiste:
I connected the modify callback in all screens where the event list drawer is used (AssociationScreen, HomeScreen and MyEventsScreen).


ntissieres (approved):
Everything is functional, thank you for your work !

PR #373, opened 2024-05-31
Add a GPS Service.

### LINKED ISSUE #318: Implement a GPS service to get the user position

alejandrocalles:
Closes #318.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -110,7 +110,10 @@ dependencies {
     // Map Libre
     implementation(libs.android.sdk)
     implementation(libs.ramani.maplibre)
-    
+
+    // Location
+    implementation(libs.play.services.location)
+
     // Hilt
     implementation(libs.hilt.android)
     implementation(libs.androidx.hilt.navigation.compose)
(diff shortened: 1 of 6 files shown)


violoncelloCH (commented):
Nice, thanks for working on this! Really clean code and a nice structure with the interface, implementation and hilt module!

I just have a concern regarding the use o the proprietary google play library. Since we so far managed to avoid dependencies on any proprietary library, this makes the app require the presence of Googles proprietary services. It will not work on devices not having them installed like Huaweis more recent Smartphone, Amazon's Android devices, Windows new capability for running Android apps as well as privacy friendly Android variants coming without Google Services by design. 
The solution for this would be to uses pure Android APIs for the location as detailed e.g. in this blog post: https://fobo66.dev/post/play-services-location-migration/#how-to-request-current-location-with-locationmanager
However given the approaching deadline, I'm not sure if it's reasonable to attempt this now. What do you think?

alejandrocalles:
It's a good point, and I hadn't really thought about it.Unfortunately, I don't think it would be reasonable to make this changes in this PR. However, I do believe it would be better to stay away from proprietary libraries, so I would consider creating an issue in case this project is continued in the future :)

srsingh04 (approved):
Nice and clean code! Good job, LGTM: ready to be merged.

violoncelloCH (approved):
as already said, nice implementation! thanks for the massive work!
I'll create an issue as discussed for the dependency concern :)

PR #374, opened 2024-05-31
bugfix to clear filters

Greenade:


ntissieres (approved):
Thanks for this fix !

PR #375, opened 2024-05-31
Fix/minor UI fixes

ntissieres:
Correct the search association by tags and layout for leave events


alejandrocalles (approved):
All looks good.

PR #379, opened 2024-05-31
Wrong call back bug

octogradiste:
The `viewOnMap` call back was set to `modify`. Now I changed it to be explicit + added a new test.


ntissieres (approved):
The problem seems corrected, thanks !

PR #381, opened 2024-06-01
Let the user view an event from the list on the map.

### LINKED ISSUE #380: Connect the list view to the bottom sheet on the map

alejandrocalles:
Should be able to click on  the map icon in an event on the Home Screen and view it on the map.

Closes #380.

### DIFF
--- a/app/src/main/java/com/github/swent/echo/compose/map/MapViewProviderImpl.kt
+++ b/app/src/main/java/com/github/swent/echo/compose/map/MapViewProviderImpl.kt
@@ -17,7 +17,7 @@ import com.mapbox.mapboxsdk.maps.MapView
 import com.mapbox.mapboxsdk.maps.MapboxMap
 import com.mapbox.mapboxsdk.maps.Style
 
-class MapLibreMapViewProvider : IMapViewProvider<MapView> {
+class MapViewProviderImpl : MapViewProvider<MapView> {
 
     // Stores the last camera position so we can restore it when the map is recreated (i.e. when the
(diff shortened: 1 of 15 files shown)


alejandrocalles:
Clicking on an event's 'view on map' button also centers the map on the event.

app/src/main/java/com/github/swent/echo/compose/map/MapViewProviderImpl.kt
```diff
@@ -78,6 +78,11 @@ class MapLibreMapViewProvider : IMapViewProvider<MapView> {
         }
     }
 
+    override fun setSavedCameraPosition(newPosition: LatLng, zoomLevel: Double) {
+        lastCameraPosition =
+            CameraPosition.Builder().target(newPosition).zoom(zoomLevel).bearing(0.0).build()
```
srsingh04, line 83: Knitpick: Isn't .bearing(0.0) redundant here? Is the default value of the bearing not 0?

srsingh04 (approved):
Thank you for working on this enhancement on the maps. It would really improve user experience of our application. 
Ready to be merged!

app/src/main/java/com/github/swent/echo/compose/map/MapViewProviderImpl.kt
alejandrocalles: You're right, I'll change that :)

srsingh04 (approved):
LGTM! Ready to be merged

PR #385, opened 2024-06-02
integrated camera feature

srsingh04:
Integrated a camera feature through which now the user can take a take a picture/selfie and upload it as their profile picture. 
Changing back the ProfileCreationviewModel back to an old version as well to fix a bug in the UI update.


alejandrocalles (changes requested):
Everything seems to work really well! Thank you :) Good job and really good addition.

A small UI comment: I believe the 'edit' overlay on the profile picture (in the profile screen) shouldn't be there. The user would expect said button to trigger the camera, but there's already a camera button that does that. That would also allow the user to see their profile picture more clearly.

Also a small note on logs. I believe they should be removed after debugging, or at least keep those that are related to relatively important errors (e.g. the user denied permissions, etc), and not those that are related to simple UI updates, for instance, like 'Last name field'. For those that are important, you could use priorities other than `d` (debugging), such as `w` (warning) or `e` (error).

srsingh04:
> I believe the 'edit' overlay on the profile picture (in the profile screen) shouldn't be there. The user would expect said button to trigger the camera, but there's already a camera button that does that.
This part of the feature has been implemented by @unglazedstamp and I think how it differs from the camera is for being solely there as a functionality to select a picture from the personal gallery of the user. Nonetheless, I personally agree with your idea and think it's actually better to display the profile picture in large and clear and maybe put the edit icon on the circumference of the profile picture circle. This can be an interesting task for a future PR!

alejandrocalles:
I understand. Is it possible to remove the camera button and connect the camera launcher to the edit button instead. It's just weird to have a non-functional button in production code.

app/src/main/java/com/github/swent/echo/compose/authentication/CreateProfile.kt
```diff
@@ -134,6 +140,8 @@ fun ProfileCreationScreen(
 
     if (state == CreateProfileState.SAVED) {
         LaunchedEffect(state) { navAction.navigateTo(Routes.MAP) }
+    } else {
+        Log.d("CreateProfile", "State: $state")
```
unglazedstamp: This log as well as all the others in this file should be removed or at least commented as they are debug logs

app/src/main/java/com/github/swent/echo/compose/authentication/CreateProfile.kt
```diff
@@ -648,10 +713,24 @@ fun PictureTransformer(
...
+                    Log.d("CreateProfile", "New picture: $newPicture")
                 }
             ) {
                 Text(stringResource(R.string.edit_event_screen_confirm))
             }
         }
     }
 }
+
+private fun handleCapturedImage(bitmap: Bitmap, setPicture: (Bitmap) -> Unit) {
```
unglazedstamp, line 696: This functions should be commented

app/src/main/java/com/github/swent/echo/compose/authentication/CreateProfile.kt
```diff
@@ -654,3 +692,18 @@ fun PictureTransformer(
...
+    // Process the captured image bitmap here
+    setPicture(bitmap)
+}
+
+private fun dispatchTakePictureIntent(cameraLauncher: ActivityResultLauncher<Intent>) {
```
unglazedstamp, line 701: This function should be commented

unglazedstamp (dismissed):
Nice feature. I found some improvements in the code. Also I agree that it doesn't make sense to have an icon button to take a picture and an edit button on the picture to get a picture from the gallery. It is preferable to either have a dialog composable to do the choice once the user press on the edit button on the picture or have two icon buttons below the picture (one for each option).

srsingh04:
>  It's just weird to have a non-functional button in production code.
The button is not completely non functional as such...it allows the user to choose amongst multiple images which they have already stored in the past, along with the ones they have clicked using the camera. The feature may come out as a bit confusing to the users in the beginning but unfortunately due to time constraints, I prefer not to touch the code furthermore in the PR. I agree that a dialog box linking it with the camera and a choice of choosing from the gallery would definitely make wayyyy more sense. 
Thank you for your feedback regardless!

Greenade (approved):
LGTM, could have been better as said by the others but due to time constraints we will merge it like this. This is still a very good addition to the app

violoncelloCH:
I squashed all the commits to remove all the "debug" and similar commit messages.

PR #389, opened 2024-06-02
Fix profile saving bug

octogradiste:
The problem was that we called the onSave callback (which starts a coroutine to save the profile) and didn't wait for it to finish before navigating to the map.


srsingh04 (approved):
Thank you for fixing the bug!

